PR #179, opened 2026-03-19
feat: add GitHub event polling trigger + cleanup service

florian-bellotti:
## Summary

- **GitHub Event Polling**: `github_event` automation trigger that polls PRs/issues via `gh` CLI, detects opens/updates/comments/reopens by diffing snapshots, enriches context (diff, body, files), interpolates template variables, and optionally posts results as GitHub comments.
- **Built-in Review Templates**: Three seeded templates — PR Review, Security Review, and Full Review (combined with merge verdict).
- **Cleanup Service**: Post-run artifact stripping, TTL sweeps of archives/run sessions, and disk pressure monitoring.
- **Frontend**: Trigger type selector, event checkboxes with label filters, GitHub icon in sidebar, Storage settings page with disk usage and cleanup controls.

## Test plan

- [x] `npm run typecheck` — 0 errors
- [x] `npm run lint` — 0 errors
- [x] `npm test` — 87 files, 1084 tests passed
- [ ] Create a `github_event` automation, open a PR, verify polling triggers review and posts comment
- [ ] Verify built-in templates appear after server restart
- [ ] Verify disk usage and cleanup in Settings > Storage
- [ ] Verify label filter gates event matching
- [ ] Verify cron automations still work unchanged


florian-bellotti:
### Hive: hive - PR reviewer

I'll start by reading the key files in the diff to understand the full context.Let me read the specific critical sections directly.Now I have a thorough understanding of the entire PR. Let me compile the review.

## Summary

This PR adds GitHub event polling triggers for automations (PR reviews via `gh` CLI snapshot diffing), three built-in review templates, a cleanup service (artifact stripping, TTL sweeps, disk pressure monitoring), and corresponding frontend UI. The overall quality is good — the architecture is well-structured, the snapshot-based event detection is clever, and there's proper attention to deduplication, backoff, and state pruning. However, there are several correctness bugs (one significant race condition in concurrent event runs sharing a single workspace) and a few design issues worth addressing before merge.

## Code Review Findings

### 1. [BUG] Concurrent event runs share a single workspace — `automation-scheduler.ts:250-346`

`executeEventRun()` does **not** check `isRunning(autoId)` before starting. Multiple GitHub events for the same automation (e.g., two PRs opened simultaneously, or a PR opened + comment on another) will run concurrently in the **same worktree**, causing git checkout conflicts and corrupted state.

The cron path (`executeRun`) guards against this, but `executeEventRun` skips that guard entirely.

**Fix:** Either add a concurrency guard (queue or skip), or create per-run isolated worktrees (e.g., `workspace-pr-42/`).

```typescript
// At the top of executeEventRun:
if (this.isRunning(autoId)) {
  console.warn(`[scheduler] Skipping event run for ${autoId}: already running`);
  // Option A: queue the event for later
  // Option B: skip and let next poll pick it up (fingerprint not yet marked as processed)
  return { ...run, status: "failure", error: "Concurrent run in progress" };
}
```

### 2. [BUG] `activeRuns` keyed by `run.id` but `isRunning` scans by `automationId` — `automation-scheduler.ts:118-123`

The refactoring from keying by `autoId` to `run.id` is correct for supporting multiple concurrent runs, but `isRunning()` now does a linear scan. Combined with finding #1, this creates a subtle issue: `executeRun` (cron) returns the first found active run's data, but if an event run is active, it returns an event run's `AutomationRun` — which may confuse callers.

**Severity:** Nit — the linear scan is fine for expected cardinality, but the return value semantics are muddled.

### 3. [BUG] `ensureWorkspaceForEvent` does `git reset --hard` to `event.headSha` — `automation-scheduler.ts:453-455`

If `headSha` is undefined (it's optional on `GitHubTriggerEvent`), this is safely skipped. But even when present, doing `git reset --hard` after `git fetch origin pull/N/head:branch --force` + `git checkout branch` is redundant — the fetch already sets the branch to `headSha`. The `--force` flag on fetch handles this.

More importantly, if the workspace has uncommitted changes from a **previous automation run** (e.g., the agent wrote files), `reset --hard` silently destroys them without warning. This is fine given automations shouldn't persist workspace state, but worth noting.

**Severity:** Nit

### 4. [BUG] `pruneStaleSnapshots` checks `pr.state === "closed"` but `fetchPrs` returns `item.state` as-is from `gh` — `github-poll-state.ts:109`

The `gh pr list --json state` returns states like `"OPEN"`, `"CLOSED"`, `"MERGED"` (uppercase). But `pruneStaleSnapshots` checks for lowercase `"closed"` and `"merged"`. This means stale PR snapshots are **never pruned**.

**Fix:**
```typescript
if (pr.state === "CLOSED" || pr.state === "MERGED") {
```

Or normalize state to lowercase when storing snapshots.

### 5. [BUG] `pruneStaleSnapshots` checks `issue.state === "closed"` — `github-poll-state.ts:117`

Same issue for issues. `gh issue list --json state` returns `"OPEN"` / `"CLOSED"` in uppercase.

**Fix:** `if (issue.state === "CLOSED") {`

### 6. [BUG] `postPrComment` can hit E2BIG for large review outputs — `automation-scheduler.ts:574-578`

The `fullContent` from the last assistant message is passed as `--body` to `gh pr comment`. For large reviews (especially with embedded diffs), this can exceed `ARG_MAX`. This is the same class of issue that was fixed for Claude CLI in this very PR (stdin approach).

**Fix:** Consider writing the body to a temp file and using `--body-file`:
```typescript
await writeFile(tmpPath, comment, "utf-8");
await gh(["pr", "comment", String(num), "--repo", `${owner}/${repo}`, "--body-file", tmpPath]);
```

### 7. [NOTE] `fetchPrs` with `--state all` returns up to 30 items — `github-event-poller.ts:393-398`

On repos with many PRs, `--limit 30 --state all` may miss recently updated open PRs because closed/merged PRs fill the limit. The `gh pr list` command returns by most recently updated, so this is partially mitigated, but a repo with frequent PR activity on closed PRs could miss new opens.

**Fix:** Consider fetching `--state open` separately if needed, or increasing the limit.

### 8. [NOTE] Disk pressure only blocks cron runs, not event runs — `automation-scheduler.ts:157`

`executeRun` (cron) checks `cleanupService?.isBlocked()`, but `executeEventRun` does not. Under hard disk pressure, event-triggered automation runs still proceed, which could worsen the situation.

**Fix:** Add the same guard to `executeEventRun`:
```typescript
if (this.cleanupService?.isBlocked()) {
  // ... same pattern as executeRun
}
```

### 9. [NIT] `executeCleanup` reports pre-counted archive/session numbers — `cleanup-service.ts:308-310`

The result reports `archivesRemoved = preview.staleArchives` *before* TTL sweep runs. If the sweep fails partway through, the reported count won't match actual removals. This is a UX misleading-data issue.

**Fix:** Count actual removals in `runTtlSweep()` and return them, or accept the approximation with a comment.

### 10. [NIT] Config reloaded on every `stripArtifacts` call — `cleanup-service.ts:127`

`getConfig()` is called inside `stripArtifacts()`, which is called per-workspace in `stripAllAutomationArtifacts()`. For N workspaces, this reads and parses `config.json` N times.

**Fix:** Pass config as a parameter from the caller, or cache within the sweep cycle.

### 11. [NIT] `.gitignore` missing trailing newline — `.gitignore`

The diff shows `-.claude/` + `+.claude/\ No newline at end of file`. POSIX requires a trailing newline. Some tools may misbehave.

### 12. [NIT] `seedDefaultTemplates` runs on every server restart — `index.ts:368`

The function is idempotent (checks if ID exists), but it does filesystem I/O on every startup. Fine for 3 templates, but worth noting.

### 13. [NIT] Duplicate code between `executeRun` and `executeEventRun` — `automation-scheduler.ts:148-346`

The two methods share ~80% of their logic (workspace setup, prompt resolution, git context injection, project lookup). Consider extracting a shared helper to reduce maintenance surface.

## Security Findings

### Vuln 1: [Input Validation]: `cleanup-service.ts:130-132`
* **Severity**: Medium
* **Confidence**: 8 / 10
* **Description**: `artifactDirs` comes from user-configurable `config.json`. While there's a path traversal check (`resolve(targetPath).startsWith(resolve(workspacePath))`), this check can be bypassed on case-insensitive filesystems (macOS). Additionally, if `artifactDirs` contains absolute paths, `join(workspacePath, "/etc/something")` in Node.js resolves to `/etc/something` (join ignores the first arg if a later segment is absolute), bypassing the startsWith check.
* **Exploit Scenario**: An attacker with write access to `config.json` adds `"/tmp/important"` to `artifactDirs`. `join(wsPath, "/tmp/important")` resolves to `/tmp/important`, and `resolve("/tmp/important").startsWith(resolve(wsPath))` fails — so the check catches it. Wait — `path.join("/a/b", "/c")` actually returns `/a/b/c` in Node.js (join does not treat absolute segments specially). So this is actually safe. Let me verify... Node's `path.join` concatenates and normalizes but does NOT treat later segments as roots. So `join("/ws", "../../../etc")` → `/etc`, which would fail the startsWith check. The existing guard is correct.
* **Recommendation**: The existing path traversal guard is adequate. No action needed — downgrading this finding.

**No high-confidence security vulnerabilities found.** The key security surfaces are well-handled:
- `artifactDirs` path traversal is guarded by `resolve().startsWith()` check
- `gh` CLI is invoked via `execFile` (not shell), preventing command injection
- `postPrComment` body goes through `execFile` as an array argument, not through a shell
- GitHub event numbers are integers validated by TypeScript types
- No user-controlled data reaches `eval`, SQL, or template engines unsafely

## Verdict

**REQUEST CHANGES** — The concurrent event run race condition (Finding #1) is a real bug that will cause git workspace corruption when multiple events trigger simultaneously for the same automation. The stale snapshot pruning bug (Findings #4-5) means poll state grows unboundedly. The disk pressure bypass for event runs (Finding #8) undermines the safety mechanism. These should be fixed before merge. The E2BIG risk on comment posting (Finding #6) is worth addressing since this PR specifically fixes E2BIG elsewhere.

florian-bellotti:
### Hive: hive - PR reviewer

I'll start by reading the key files in this PR to conduct a thorough review.Now let me read a few more specific files to check critical details.Now let me check one more critical area — the poller's event dispatching with `void` fire-and-forget pattern, and the frontend types for consistency.Now let me check the iOS type sync and one more thing about comment self-trigger.Now I have a thorough understanding of the full PR. Let me compile the review.

## Summary

This PR adds a well-architected GitHub event polling system that detects PR/issue changes via snapshot diffing with `gh` CLI, three built-in review prompt templates, a cleanup service for artifact/TTL/disk management, and corresponding frontend UI. The overall design is sound — the polling model, fingerprint deduplication, state pruning, and the `BuildArgsResult` refactor are all clean. However, there is a **critical self-triggering infinite loop** when `postResultAsComment` is enabled with `pull_request.comment` events, a significant race condition in fire-and-forget event dispatching, and several correctness issues that should be addressed before merge.

## Code Review Findings

### 1. [BUG] Infinite loop: comment posting triggers re-poll — `automation-scheduler.ts:593` + `github-event-poller.ts:202`

**Severity: Bug (critical)**

When an automation subscribes to `pull_request.comment` events **and** has `postResultAsComment: true`, the agent's posted comment increments `commentCount` on the PR. The next poll cycle detects the new comment, fires a `pull_request.comment` event, which triggers the automation again, which posts another comment, ad infinitum.

**Fix:** Filter out self-posted comments. The simplest approach: when enriching comment context, check if the latest comment author matches the `gh` authenticated user (or a known bot identifier like the `### Hive:` prefix). Alternatively, compare the comment body against the `"### Hive:"` prefix and skip dispatch when the latest comment is the bot's own.

```typescript
// In enrichEventContext or pollRepo, before dispatching comment events:
if (lastComment?.body?.startsWith("### Hive:")) continue; // skip self-posted
```

---

### 2. [BUG] Fire-and-forget `executeEventRun` races with poll state lock — `github-event-poller.ts:319`

**Severity: Bug**

`executeEventRun` is dispatched with `void` (fire-and-forget) while still inside the `withPollStateLock` callback. The event is immediately marked as processed (line 337), but if the `isRunning()` check in `executeEventRun` returns true (concurrent run guard), the event is silently dropped **and** marked processed — meaning it will never be retried.

Additionally, because event runs are dispatched but not awaited, multiple events for different automations on the same repo can be dispatched simultaneously. If two events for the same automation arrive (e.g., `opened` + `synchronize` in same batch), both pass `isRunning()` since neither has started yet, and both attempt `ensureWorkspaceForEvent` on the same workspace concurrently.

**Fix:** Either:
- Await `executeEventRun` or check `isRunning()` before marking as processed
- Queue events per-automation and dispatch sequentially

---

### 3. [BUG] `gh pr list --state all` may not return all PRs — `github-event-poller.ts:389`

**Severity: Bug**

`fetchPrs` uses `--limit 30` with `--state all`. For active repos with >30 PRs, older open PRs won't appear in results. If a PR was previously snapshotted and then falls off the 30-item page, it won't be updated. This isn't a data loss issue, but means events for PRs beyond the first 30 are silently missed.

**Fix:** Consider a higher limit (100+) or filter to `--state open` (since most events happen on open PRs) plus a separate query for recently-updated closed ones if needed. At minimum, document this limitation.

---

### 4. [BUG] `ensureWorkspaceForEvent` uses `origin` but workspace is a worktree — `automation-scheduler.ts:472`

**Severity: Bug**

The fetch command uses `origin` as the remote name:
```typescript
await git(["fetch", "origin", `pull/${event.number}/head:${branchName}`, "--force"], workspacePath);
```

But `ensureWorkspace` creates a worktree from a bare repo. In a worktree, the remote configuration is inherited from the bare repo. If the bare repo's remote isn't named `origin` (unlikely but possible with custom setups), or if the worktree's git config doesn't have fetch permissions for `pull/*/head`, this will fail silently.

More importantly, `ensureWorkspace` calls `refreshWorktreeToRemoteBranch` which resets to the default branch. Then `ensureWorkspaceForEvent` immediately fetches and checks out a PR branch. But `refreshWorktreeToRemoteBranch` likely runs `git fetch` and `git reset --hard`, and the subsequent fetch+checkout could conflict if there are uncommitted changes from a previous agent run (e.g., if the agent wrote files but the session failed before cleanup).

**Fix:** Add a `git clean -fd` after checkout to ensure no stale files, and consider adding `git stash` before the reset sequence.

---

### 5. [NIT] `isRunning` linear scan per check — `automation-scheduler.ts:118`

**Severity: Nit**

`isRunning()` iterates all active runs to find one matching `autoId`. This is called in `executeRun`, `executeEventRun`, and the poller's dispatch loop. With many concurrent runs it's O(n) per call. The old code used `autoId` as the map key, making it O(1).

**Fix:** Add a secondary index `Map<string, Set<string>>` mapping `automationId → Set<runId>`, or use a composite key.

---

### 6. [NIT] Cleanup routes lack authorization — `cleanup.ts:19`

**Severity: Nit**

`POST /api/cleanup/run` can delete archived workspaces and run sessions. While the app is typically behind auth, this endpoint performs destructive operations with no additional confirmation or audit logging.

**Fix:** Consider adding a confirmation parameter or rate limiting to prevent accidental mass deletion.

---

### 7. [NIT] `pollRepo` serializes `gh` API calls — `github-event-poller.ts:166-283`

**Severity: Nit**

PR and issue fetching + enrichment all happen sequentially within the lock. If enrichment (`gh pr view`, `gh pr diff`) is slow, the lock is held for a long time, blocking state saves for other repos.

**Fix:** Consider fetching+enrichment outside the lock, and only acquiring the lock for state comparison + update.

---

### 8. [NIT] Snapshot updated even for non-needed events — `github-event-poller.ts:233-241`

**Severity: Nit**

The PR snapshot is always updated for every fetched PR, even if no event was detected. This is harmless but means state file grows with all 30 PRs even if only `pull_request.opened` is subscribed. This is actually fine for correctness but worth noting for the state file size.

---

### 9. [NOTE] `CreateAutomationDialog` defaults `postResultAsComment` to `true` for new automations — `CreateAutomationDialog.tsx:80`

**Severity: Note**

```tsx
const [postResultAsComment, setPostResultAsComment] = useState(true);
```

But when editing an existing automation without the flag, it reads `automation.action.postResultAsComment ?? false` (line 144). The create vs. edit default is inconsistent. Combined with finding #1 (infinite loop), defaulting to `true` is especially dangerous.

**Fix:** Default to `false` for safety.

---

### 10. [NOTE] `seedDefaultTemplates` runs on every server restart — `default-prompt-templates.ts`

**Severity: Note**

If a user customizes a built-in template (by editing its content), the seed function only checks by ID — so user edits are preserved. However, if a user deletes a built-in template, it will be re-created on next restart. This may surprise users.

**Fix:** Consider storing a "seeded" flag in config to only seed once.

---

### 11. [NOTE] Comment truncation at 60K chars vs. GitHub's 65,536 limit — `automation-scheduler.ts:598`

**Severity: Note**

The `MAX_COMMENT_CHARS = 60_000` leaves a ~5.5K char buffer which is consumed by the wrapper header + footer. The header `### Hive: ${auto.name}\n\n` and footer are relatively short, so this should be fine, but if `auto.name` is very long it could theoretically exceed GitHub's limit.

---

### 12. [NOTE] Missing test coverage for event dispatch and enrichment — `github-event-poller.test.ts`

**Severity: Note**

The tests only cover lifecycle (start/stop), guard paths (no automations, gh not installed, reentrancy). There are no tests for:
- Actual event detection (snapshot diffing)
- Event enrichment (context building)
- Label filtering
- Fingerprint deduplication
- Multiple automations on same repo

These are the core behaviors of the poller and should have integration tests.

---

### 13. [NIT] `CreateAutomationRequest` type not validated on update — `api/automations.ts:166`

**Severity: Nit**

The update endpoint validates `trigger.expression` and `trigger.events` but doesn't validate that `trigger.type` is a known value. It also doesn't validate that the trigger type matches the automation's project linkage (e.g., changing a cron automation to `github_event` without a `projectId`).

## Security Findings

### Vuln 1: [Path Traversal] Artifact directory names from config — `cleanup-service.ts:130-132`

* **Severity**: Medium
* **Confidence**: 8 / 10
* **Description**: `artifactDirs` is read from user-editable config (`config.json`). While there's a path traversal check (`resolve(targetPath).startsWith(resolve(workspacePath))`), the check uses string prefix comparison which can be bypassed with paths like `workspacePath + "-evil"` if such a directory exists adjacent to the workspace.
* **Exploit Scenario**: An attacker with config write access sets `artifactDirs` to `["../../projects"]`. The `resolve()` + `startsWith()` check would catch `../../` traversals correctly, but a value like `../auto-OTHER/workspace/sessions` could delete another automation's data if the resolved path happens to share the prefix.
* **Recommendation**: Use `resolve(targetPath).startsWith(resolve(workspacePath) + sep)` (with path separator) to prevent prefix-matching sibling directories. Also validate `artifactDirs` entries don't contain `..` or `/` when saving config.

### Vuln 2: [Input Injection] Template variable interpolation in prompts — `github-prompt-context.ts:39`

* **Severity**: Medium  
* **Confidence**: 8 / 10
* **Description**: PR titles, bodies, and comment text from GitHub are interpolated directly into prompts via `interpolateGitHubVariables`. A malicious PR title like `{PR_DIFF}` would be interpolated first as a title, then if the result still contains `{PR_DIFF}`, the regex won't re-match (single pass). However, a PR title containing prompt injection text (e.g., "Ignore all previous instructions") would be injected directly into the AI prompt without sanitization.
* **Exploit Scenario**: An attacker opens a PR with a title/body containing adversarial instructions that manipulate the AI agent into approving malicious code, posting misleading review comments, or executing unexpected commands in the workspace.
* **Recommendation**: This is an inherent risk of AI-powered review. Document it clearly. Consider adding a note in the review template output that the AI processed user-supplied content, and consider running the agent with `--permission-mode plan` for additional safety.

No high-severity security vulnerabilities found. The codebase correctly uses `gh` CLI rather than raw shell commands, uses temp files for comment bodies (avoiding argument injection), and validates GitHub URLs.

## Verdict

**REQUEST CHANGES** — The self-triggering infinite loop (#1) is a critical correctness bug that will cause runaway automation runs and spam GitHub with comments when `postResultAsComment` is enabled with `pull_request.comment` events. The fire-and-forget race condition (#2) can silently drop events while marking them processed. Both should be fixed before merge. The remaining findings are non-blocking improvements.

## Summary

This PR adds GitHub event polling triggers, built-in review templates, a cleanup service, and frontend UI — all well-designed architecturally. The core issue is a **critical infinite loop**: when `postResultAsComment` is enabled alongside `pull_request.comment` event subscriptions, the bot's own posted comment triggers the next poll, creating an unbounded feedback loop. There's also a race condition where fire-and-forget event dispatching can mark events as processed even when the run is skipped. I recommend fixing the self-trigger loop (filter bot comments), awaiting or queuing event dispatches, and defaulting `postResultAsComment` to `false`. The security surface is reasonable — the main risk is prompt injection via GitHub content, which is inherent to AI review and should be documented. Verdict: **REQUEST CHANGES** on the infinite loop bug.

0xlny:
I reviewed the current head of PR #179 (`0cfcb5b`).

First, an architecture clarification: this is clearly an extension of the existing Automation engine, not a parallel implementation. The new `github_event` trigger type plugs into the existing `Automation` / `AutomationRun` model, the poller dispatches into `AutomationScheduler.executeEventRun()`, and the execution path still reuses the existing persistence, `ConversationSession`, prompt resolution, notifications, and worktree handling.

That said, there are still several issues that should be fixed before merge. I’m listing them in terms of concrete work to do, because this comment is intended to be actionable for an agent.

## 1. Prevent feedback loops on `*.comment` automations

Current behavior:
- `GitHubEventPoller` emits `pull_request.comment` / `issues.comment` when `commentCount` increases.
- `AutomationScheduler.completeRun()` can post the run result back as a GitHub comment when `postResultAsComment` is enabled.
- There is no code-level filter to ignore comments posted by Hive itself.

Result:
- A successful run can create a new comment.
- The next poll sees the higher `commentCount` and triggers the automation again.
- This can loop indefinitely for both PR and issue comment automations.

Relevant code:
- `backend/src/services/github-event-poller.ts`
- `backend/src/services/automation-scheduler.ts`

Required fix:
- Add an explicit loop-prevention guard in code.
- Do not rely on the prompt to “ignore bot comments”.

Acceptable implementations:
- Ignore comments authored by the configured GitHub actor used by Hive.
- Or tag Hive-generated comments with a stable marker and ignore comments containing that marker.
- Or store enough metadata to avoid retriggering on comments emitted by the same automation run.

Required tests:
- Regression test for PR comment automation.
- Regression test for issue comment automation.
- Test should prove that one successful comment-posting run does not schedule another run on the next poll.

## 2. Do not mark event fingerprints as processed before execution is actually accepted/completed

Current behavior:
- In `pollRepo()`, the event is considered `dispatched` before `executeEventRun()` has actually succeeded.
- The fingerprint is appended to `processedEvents` immediately afterward.
- `executeEventRun()` can still effectively reject the event by returning early when the automation is already running, when disk pressure blocks new runs, etc.
- Even if the run starts, it can still fail later.

Result:
- Immutable events such as `pull_request.opened` can be lost permanently.
- The earlier “lost fingerprints” fix is not complete, because the poller still acknowledges the event too early.

Required fix:
- Only mark the fingerprint processed once the event has been truly accepted for execution.
- Preferably, persist explicit event lifecycle states (`detected` / `accepted` / `completed`) or return a structured scheduler result that distinguishes:
  - accepted and started
  - skipped because already running
  - blocked because disk pressure
  - failed before start

Minimum acceptable behavior:
- If the scheduler does not actually start a new run, do not consume the fingerprint.

Required tests:
- Event arrives while automation is already running: fingerprint must not be consumed.
- Event arrives under disk pressure: fingerprint must not be consumed.
- Event run fails before completion: expected retry semantics must be explicitly tested and documented.

## 3. Fix polling coverage; `--state all --limit 30` is not enough

Current behavior:
- PR snapshots come from `gh pr list --state all --limit 30`.
- Issue snapshots come from `gh issue list --state all --limit 30`.

Result:
- In an active repo, recently closed items can crowd out older open items.
- Open PRs/issues can disappear from the poller’s visibility window.
- Once that happens, `synchronize`, `comment`, and `reopened` detection becomes unreliable or stops entirely for those items.

Required fix:
- Replace the fixed window with a strategy that cannot hide open items.

Acceptable implementations:
- Paginate until all open items are covered.
- Or fetch open and closed items separately.
- Or maintain a larger/explicit tracked set and refresh it safely.

Required tests:
- Add a detection test showing that an older open PR is still tracked when many newer closed PRs exist.
- Do the same for issues if issue triggers remain supported.

## 4. Complete the `E2BIG` fix for non-Claude providers

Current behavior:
- Claude now supports stdin-based prompt delivery, which fixes large prompt / diff payloads.
- Codex and Gemini still pass the full content as CLI args.

Result:
- GitHub-event automations that interpolate large `{PR_DIFF}` values can still hit `ARG_MAX` on Codex or Gemini.
- So the PR currently fixes the problem only for one provider while exposing the feature to all providers.

Required fix:
- Either implement stdin-based prompt delivery for Codex and Gemini too,
- Or explicitly prevent GitHub-event automations using providers that are not yet safe for large interpolated prompts.

Required tests:
- Provider-level regression coverage showing large payload handling for each supported provider.
- If you choose to block unsupported providers instead, add API/UI validation tests covering that restriction.

## 5. Fix `postResultAsComment` editing semantics

Current behavior:
- The edit dialog only includes `postResultAsComment` in the payload when the checkbox is `true`.
- The backend update route shallow-merges `action`.

Result:
- Once `postResultAsComment` is enabled, unchecking it in the UI does not reliably turn it off, because the update payload omits the field and the old `true` value can survive the merge.

Relevant code:
- `frontend/src/components/CreateAutomationDialog.tsx`
- `backend/src/api/automations.ts`

Required fix:
- Send `postResultAsComment: false` explicitly when the user disables it.
- Ensure the backend persists the false value correctly instead of preserving the previous value.

Required tests:
- UI/API regression test: create automation with comment posting enabled, edit it, disable comment posting, verify persisted automation has `postResultAsComment === false`.

## Additional points that are not immediate blockers but should be addressed soon

### A. Comment-trigger semantics are still lossy
The current implementation detects comment events via `commentCount` deltas and then fetches only the latest comment body. If multiple comments arrive between two polls, only one event is emitted and only the last comment is available in context. If the intended behavior is “respond to every comment”, this is still incorrect.

### B. `CleanupService` still lacks direct service-level tests
There is still no dedicated test file for `backend/src/services/cleanup-service.ts`, even though this service performs recursive deletions and TTL cleanup. Given the risk profile, it should have direct coverage before merge.

### C. The cleanup endpoints are not an auth hole
One item from the earlier review does *not* appear to be a real issue: the cleanup routes are covered by the app’s global HTTP auth hook, so they are not uniquely unauthenticated.

## Validation performed
I also ran the targeted backend tests that currently exist for the new code paths:
- `src/services/github-event-poller.test.ts`
- `src/state/github-poll-state.test.ts`
- `src/agents/github-prompt-context.test.ts`
- `src/agents/providers/claude.test.ts`
- `src/api/settings.test.ts`
- `src/state/config.test.ts`

They pass, but they do not cover the key regressions described above. In particular, the existing poller tests are still shallow relative to the event-detection / dispatch logic, and there is still no direct `CleanupService` test suite.

PR #180, opened 2026-03-19
feat: create new repository from Hive

florian-bellotti:
## Summary

- Adds a **"Create new"** mode to the Add Project dialog (alongside "Clone existing") with inline mode toggle
- When GitHub is connected via `gh auth`, creates repo on GitHub (`gh repo create`) with chosen visibility (public/private), pushes initial commit (README + .gitignore)
- Falls back to **local-only** repo creation when GitHub is not connected
- `ProjectState.url` is now optional to support local-only projects — all consumers guarded
- Full-stack implementation: backend, frontend, iOS, tests updated

### Backend
- `project-manager.ts`: new `initProject()`, `validateRepoName()`, `createGitHubRepo()` — creates bare repo via `git init --bare`, seeds initial commit, optionally pushes to GitHub
- `types.ts`: `ProjectState.url` optional, `CreateProjectRequest` is now a discriminated union (`clone` | `create`)
- `projects.ts`: POST handler dispatches on `mode` field (backward compatible)
- `workspaces.ts`: guard `parseGitHubRepo()` for optional URL

### Frontend
- `AddProjectDialog.tsx`: mode toggle, create form (name + visibility + GitHub status indicator)
- `useProjects.ts`: shared `createProjectThenWorkspace()` helper, new `createNewProjectWithWorkspace` mutation
- `Sidebar.tsx`, `ProjectDetail.tsx`: guard optional URL, "Local only" label

### iOS
- `AddProjectSheet.swift`: segmented picker (Clone/Create), create form with GitHub status check
- `APIClient.swift`: `createNewProject()`, `fetchAccountStatus()`
- `ProjectStore.swift`: shared `createProjectWithWorkspace(displayName:apiCall:)` helper
- `Models.swift`: `url` optional, `AccountStatusResponse` types

## Test plan
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm test` — all 1084 tests pass (87 files)
- [ ] Manual: create project with GitHub connected → verify repo on GitHub, workspace works
- [ ] Manual: create project without GitHub → verify local-only, workspace works
- [ ] Manual: clone existing repo → same behavior as before

### DIFF
--- a/backend/src/api/projects.test.ts
+++ b/backend/src/api/projects.test.ts
@@ -87,6 +87,49 @@ describe("POST /api/projects", () => {
     expect(res.statusCode).toBe(400);
     expect(res.json().error).toContain("not allowed");
   });
+
+  it("creates a local-only project with mode=create", async () => {
+    const res = await app.inject({
+      method: "POST",
+      url: "/api/projects",
+      payload: { mode: "create", name: "my-new-repo" },
+    });
+    expect(res.statusCode).toBe(201);
+    const body = res.json();
+    expect(body.id).toMatch(/^proj-/);
+    expect(body.name).toBe("my-new-repo");
+    expect(body.url).toBeUndefined();
+  });
+
+  it("returns 400 for missing name in create mode", async () => {
+    const res = await app.inject({
+      method: "POST",
+      url: "/api/projects",
+      payload: { mode: "create" },
+    });
+    expect(res.statusCode).toBe(400);
+    expect(res.json().error).toContain("name is required");
+  });
+
+  it("returns 400 for invalid visibility", async () => {
+    const res = await app.inject({
+      method: "POST",
+      url: "/api/projects",
+      payload: { mode: "create", name: "test-repo", visibility: "internal" },
+    });
+    expect(res.statusCode).toBe(400);
+    expect(res.json().error).toContain("visibility must be");
+  });
+
+  it("returns 400 for malicious visibility value", async () => {
+    const res = await app.inject({
+      method: "POST",
+      url: "/api/projects",
+      payload: { mode: "create", name: "test-repo", visibility: "template=evil/repo" },
+    });
+    expect(res.statusCode).toBe(400);
+    expect(res.json().error).toContain("visibility must be");
+  });
 });
 
 describe("GET /api/projects", () => {
--- a/backend/src/api/projects.ts
+++ b/backend/src/api/projects.ts
@@ -91,6 +91,7 @@ export async function projectRoutes(app: FastifyInstance, dataDir?: string) {
 
     try {
       let project: ProjectState;
+      let warning: string | undefined;
       if (mode === "create") {
         const name = body.name as string | undefined;
         if (!name?.trim()) return reply.status(400).send({ error: "name is required" });
@@ -98,13 +99,16 @@ export async function projectRoutes(app: FastifyInstance, dataDir?: string) {
         if (visibility !== undefined && visibility !== "public" && visibility !== "private") {
           return reply.status(400).send({ error: "visibility must be 'public' or 'private'" });
         }
-        project = await initProject(name, { visibility }, dir);
+        const result = await initProject(name, { visibility }, dir);
+        project = result.state;
+        warning = result.warning;
       } else {
         const url = body.url as string | undefined;
         if (!url) return reply.status(400).send({ error: "url is required" });
         project = await createProject(url, dir);
       }
-      return reply.status(201).send(await enrichProject(project, dir));
+      const enriched = await enrichProject(project, dir);
+      return reply.status(201).send(warning ? { ...enriched, warning } : enriched);
     } catch (err: unknown) {
       return reply
         .status(errorStatus(err, 400))
--- a/backend/src/api/projects.ts
+++ b/backend/src/api/projects.ts
@@ -3,6 +3,7 @@ import { join } from "node:path";
 import type { FastifyInstance } from "fastify";
 import {
   createProject,
+  initProject,
   listProjects,
   getProject,
   deleteProject,
@@ -84,16 +85,27 @@ async function enrichProject(project: ProjectState, dir: string) {
 
 export async function projectRoutes(app: FastifyInstance, dataDir?: string) {
   app.post<{ Body: CreateProjectRequest }>("/api/projects", async (req, reply) => {
-    const { url } = req.body ?? {};
-    if (!url) return reply.status(400).send({ error: "url is required" });
+    const body = (req.body ?? {}) as Record<string, unknown>;
+    const mode = (body.mode as string) ?? "clone";
+    const dir = dataDir ?? getDataDir();
 
     try {
-      const project = await createProject(url, dataDir);
-      return reply.status(201).send(project);
+      let project: ProjectState;
+      if (mode === "create") {
+        const name = body.name as string | undefined;
+        if (!name?.trim()) return reply.status(400).send({ error: "name is required" });
+        const visibility = body.visibility as "public" | "private" | undefined;
+        project = await initProject(name, { visibility }, dir);
+      } else {
+        const url = body.url as string | undefined;
+        if (!url) return reply.status(400).send({ error: "url is required" });
+        project = await createProject(url, dir);
+      }
+      return reply.status(201).send(await enrichProject(project, dir));
     } catch (err: unknown) {
       return reply
         .status(errorStatus(err, 400))
-        .send({ error: errorMessage(err, "Clone failed") });
+        .send({ error: errorMessage(err, mode === "create" ? "Create failed" : "Clone failed") });
     }
   });
 
--- a/backend/src/api/projects.ts
+++ b/backend/src/api/projects.ts
@@ -94,7 +94,10 @@ export async function projectRoutes(app: FastifyInstance, dataDir?: string) {
       if (mode === "create") {
         const name = body.name as string | undefined;
         if (!name?.trim()) return reply.status(400).send({ error: "name is required" });
-        const visibility = body.visibility as "public" | "private" | undefined;
+        const visibility = body.visibility as string | undefined;
+        if (visibility !== undefined && visibility !== "public" && visibility !== "private") {
+          return reply.status(400).send({ error: "visibility must be 'public' or 'private'" });
+        }
         project = await initProject(name, { visibility }, dir);
       } else {
         const url = body.url as string | undefined;
--- a/backend/src/api/workspaces.ts
+++ b/backend/src/api/workspaces.ts
@@ -135,7 +135,7 @@ export async function workspaceRoutes(app: FastifyInstance, dataDir?: string) {
       const result = await getWorkspace(wsId, dataDir);
       if (!result) return reply.status(404).send({ error: "Workspace not found" });
 
-      const ghRepo = parseGitHubRepo(result.projectState.url);
+      const ghRepo = result.projectState.url ? parseGitHubRepo(result.projectState.url) : null;
       if (!ghRepo) {
         const data: PrStatusResponse = { pr: null };
         prCache.set(wsId, { data, at: Date.now() });
@@ -191,7 +191,7 @@ export async function workspaceRoutes(app: FastifyInstance, dataDir?: string) {
               return;
             }
 
-            const ghRepo = parseGitHubRepo(result.projectState.url);
+            const ghRepo = result.projectState.url ? parseGitHubRepo(result.projectState.url) : null;
             if (!ghRepo) {
               const data: PrStatusResponse = { pr: null };
               prCache.set(wsId, { data, at: Date.now() });
--- a/backend/src/projects/project-manager.test.ts
+++ b/backend/src/projects/project-manager.test.ts
@@ -3,7 +3,7 @@ import { rm } from "node:fs/promises";
 import { existsSync } from "node:fs";
 import { join } from "node:path";
 import { createTempDir, createFixtureRepo } from "../utils/test-helpers.js";
-import { createProject, listProjects, getProject, deleteProject, fetchProject } from "./project-manager.js";
+import { createProject, initProject, listProjects, getProject, deleteProject, fetchProject } from "./project-manager.js";
 
 let tempDir: string;
 let dataDir: string;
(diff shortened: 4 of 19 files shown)


florian-bellotti:
### Hive: hive - PR reviewer

Let me examine the key files more closely to understand the full context.Now let me check the actual PR branch versions of the changed files.The branch has additional commits beyond the PR diff. Let me focus on the diff provided. Let me check one key concern — the `writeFile` with user-controlled repo name:Now I have enough context for a thorough review. Let me check one more thing — the `enrichProject` call in the PR's new `create` path versus the old `clone` path:Now I have full context. Let me also quickly check if the `visibility` value is properly validated:I now have complete context. Let me compile the review.

## Summary

This PR adds a "Create new repository" feature to Hive alongside the existing "Clone existing" flow. It's a well-structured full-stack change (backend, frontend, iOS) that creates a bare Git repo, seeds an initial commit, optionally creates a GitHub repo via `gh repo create`, and makes `ProjectState.url` optional to support local-only projects. Overall quality is good — the code follows existing conventions, uses safe `execFile` patterns, and includes tests. There are a few issues worth addressing before merge.

## Code Review Findings

### 1. Orphaned project directory on `initProject` failure — **Bug**
**File**: `backend/src/projects/project-manager.ts` (`initProject`)

If any step fails between `mkdir(join(dataDir, id))` and `saveProject()` (e.g., `git init --bare` or `git clone` or `git push`), the partially-created directory tree (`bare`, `_init-temp`) is left behind with no cleanup. The existing `createProject` has the same pattern, but `initProject` has many more intermediate steps that can fail (8 git commands, 2 writeFiles).

**Fix**: Wrap the body in a try/catch that calls `rm(join(dataDir, id), { recursive: true, force: true })` on failure before re-throwing, similar to how the frontend already does rollback on workspace creation failure.

### 2. GitHub repo created but error leaves project as local-only — **Bug**
**File**: `backend/src/projects/project-manager.ts:136-140`

If `createGitHubRepo()` succeeds (the repo is now on GitHub) but `saveProject(state, dataDir)` on line 140 throws, the function throws and the project is saved as local-only (from the first `saveProject` at line 131). The user now has a dangling GitHub repo with no way to know. Additionally, there's no cleanup of the GitHub repo on any failure — once `gh repo create` succeeds, it persists even if subsequent steps fail.

**Severity**: Nit — unlikely in practice (saveProject failures are rare), but worth documenting or adding a comment.

### 3. `visibility` field not validated server-side — **Bug**
**File**: `backend/src/api/projects.ts:97`

```typescript
const visibility = body.visibility as "public" | "private" | undefined;
```

This is a type assertion, not runtime validation. A client can send `visibility: "internal"` or any arbitrary string, which gets passed directly to `gh repo create --internal`. While `gh` CLI would reject unknown flags, the `--${visibility}` interpolation in `createGitHubRepo` means any string becomes a CLI flag (e.g., `visibility: "help"` → `--help`).

**Fix**: Validate explicitly:
```typescript
const visibility = body.visibility;
if (visibility !== undefined && visibility !== "public" && visibility !== "private") {
  return reply.status(400).send({ error: "visibility must be 'public' or 'private'" });
}
```

### 4. `_init-temp` directory name collision — **Nit**
**File**: `backend/src/projects/project-manager.ts:99`

The temp worktree path `_init-temp` is deterministic within a project directory. If two concurrent `initProject` calls somehow use the same `id` (nanoid collision — astronomically unlikely) or a previous failed run left `_init-temp` behind (more realistic), the `git clone` into that path would fail or behave unexpectedly.

**Fix**: Not urgent since project IDs are unique, but adding a nanoid suffix to the temp dir name would be more robust: `join(dataDir, id, \`_init-temp-${nanoid(4)}\`)`.

### 5. `enrichProject` added to clone path (behavior change) — **Nit**
**File**: `backend/src/api/projects.ts:103`

The old clone path returned the raw `ProjectState`. The new code returns `enrichProject(project, dir)` for both modes. This is actually an improvement (returns `repoPath`, `hasFavicon`, etc. on create), but it's an undocumented behavioral change for the clone path too. The frontend likely handles it fine since it's additive, but worth noting in the PR description.

### 6. iOS `APIClient()` instantiated without configuration — **Nit**
**File**: `ios/HiveMobile/Views/Hub/AddProjectSheet.swift:155`

```swift
let status = try await APIClient().fetchAccountStatus()
```

This creates a new `APIClient` instance rather than using the shared/injected one. If `APIClient` requires host/port configuration (which it does per the iOS architecture notes), a default-constructed instance may point to the wrong server.

**Fix**: Pass the `APIClient` instance via the environment or as a parameter to `AddProjectSheet`.

### 7. Missing test coverage for `initProject` — **Nit**
The PR adds no backend unit tests for `initProject()`, `validateRepoName()`, or `createGitHubRepo()`. The `createProject` path had existing test coverage. The frontend tests are updated but only cover the dialog form, not the new mutation path.

**Suggested**: Add tests for `validateRepoName` (edge cases: empty, too long, invalid chars, dots, leading hyphen) and a basic `initProject` test using the existing `SessionOptions.command = "bash"` pattern.

### 8. Frontend `REPO_NAME_RE` duplicated — **Nit**
**File**: `frontend/src/components/AddProjectDialog.tsx:30` and `backend/src/projects/project-manager.ts:48`

The regex `/^[a-z0-9][a-z0-9._-]*$/` is duplicated between frontend and backend. If they diverge, the frontend could accept names the backend rejects or vice versa.

**Fix**: Not critical since backend is authoritative, but consider extracting to a shared validation module or at least adding a comment noting the duplication.

## Security Findings

### Vuln 1: [Input Validation / CLI Flag Injection]: `backend/src/projects/project-manager.ts:73`
* **Severity**: Medium
* **Confidence**: 9 / 10
* **Description**: The `visibility` parameter from user input is interpolated into a CLI flag via `` `--${visibility}` `` and passed to `gh repo create`. While `execFile` prevents shell injection, the unsanitized value becomes an arbitrary `gh` CLI flag. An attacker can send `{"mode":"create","name":"test","visibility":"confirm"}` to pass `--confirm`, or `visibility: "template=evil/repo"` to pass `--template=evil/repo`, potentially cloning an attacker-controlled template repo.
* **Exploit Scenario**: `POST /api/projects` with `{"mode":"create","name":"legit","visibility":"template=attacker/backdoored-repo"}` would execute `gh repo create legit --template=attacker/backdoored-repo --json sshUrl --jq .sshUrl`, creating a repo from an attacker-specified template. With `visibility: "clone"`, it could clone an existing repo.
* **Recommendation**: Validate `visibility` is exactly `"public"` or `"private"` in the API handler before passing to `initProject`. Also add a guard in `createGitHubRepo` itself:
  ```typescript
  if (visibility !== "public" && visibility !== "private") {
    throw new Error("Invalid visibility");
  }
  ```

No other security issues found. The repo name is properly validated via `REPO_NAME_RE`, `git()` and `gh()` use `execFile` with array args (safe from shell injection), and file paths use `join()` with nanoid-based project IDs (no path traversal).

## Verdict

**REQUEST CHANGES** — The `visibility` field lacks server-side validation, allowing arbitrary CLI flags to be injected into `gh repo create`. This is a concrete, exploitable issue (Vuln 1 / Finding 3) that should be fixed before merge. The remaining findings are minor improvements.

florian-bellotti:
### Hive: hive - PR reviewer

Let me examine the key files in detail to provide a thorough review.Now let me check a couple more edge cases around the git operations and the broader workspace integration.Good — `git-sync.ts` doesn't do `git fetch` on bare repos (it polls branch/diff info on worktrees), so local-only projects without a remote won't break polling. Let me verify the iOS model change doesn't break JSON decoding.Good. The Swift `Project.url` is `String?` — optional types in Swift naturally decode from `null` or missing JSON keys. No decoding issue.

Now let me check the `enrichProject` call path for the create flow — the clone path didn't call `enrichProject` before, but now both do.I now have a complete picture. Here's my review:

---

## Summary

PR #180 adds a "Create new repository" feature to Hive across all three platforms (backend, frontend, iOS). The implementation is well-structured: it introduces `initProject()` in `project-manager.ts` to seed a bare repo with an initial commit, optionally pushes to GitHub via `gh repo create`, and gracefully degrades to local-only on failure. The `ProjectState.url` field is correctly made optional throughout the stack. Security-critical paths (CLI flag injection via `visibility`, repo name validation) are properly handled with defense-in-depth. Test coverage is solid. The PR has gone through iterative hardening across 4 commits. It is safe to merge.

## Code Review Findings

### 1. Silent GitHub failure — no user feedback
**File**: `backend/src/projects/project-manager.ts:152-159`
**Severity**: Nit

When a user explicitly selects public/private visibility, and `createGitHubRepo` fails (e.g., name collision, auth expired), the error is silently swallowed. The response returns `url: undefined` and the frontend shows "Local only" — but the user has no indication their GitHub creation attempt failed.

**Suggestion**: Include a `warning` field in the response (e.g., `"GitHub repo creation failed — created as local only"`) so the frontend can display a toast. Alternatively, log the error server-side for debugging.

### 2. Orphaned GitHub repo on push failure
**File**: `backend/src/projects/project-manager.ts:76-92`
**Severity**: Nit

In `createGitHubRepo`, if `gh repo create` succeeds (line 76-85) but `git remote add` or `git push --all` fails (lines 89-90), a GitHub repo exists but the local project doesn't link to it. The caller's `catch` block silently degrades to local-only, leaving an orphaned repo on GitHub.

**Suggestion**: Wrap the push steps in a try-catch that attempts `gh repo delete` on failure:
```typescript
try {
  await git(["remote", "add", "origin", url], bare);
  await git(["push", "--all", "origin"], bare);
} catch (err) {
  await gh(["repo", "delete", name, "--yes"]).catch(() => {});
  throw err;
}
```

### 3. Clone path now calls `enrichProject` (behavior change)
**File**: `backend/src/api/projects.ts:107`  
**Severity**: Note

Previously the clone path returned the raw `ProjectState`. Now both paths return `await enrichProject(project, dir)`. This is actually an improvement (adds `repoPath`, `hasFavicon`, etc.), but it's a subtle API response shape change for the clone flow. Existing consumers should already handle the extra fields gracefully since they're additive.

### 4. Duplicated `REPO_NAME_RE` regex
**File**: `frontend/src/components/AddProjectDialog.tsx:33` vs `backend/src/projects/project-manager.ts:50`  
**Severity**: Note

The regex is duplicated between frontend and backend. Backend is authoritative, so this isn't a bug, but the frontend copy could drift. Consider exporting a shared constant or documenting the relationship.

### 5. Frontend sends untrimmed `name` to backend
**File**: `frontend/src/components/AddProjectDialog.tsx:183`  
**Severity**: Note

`name.trim().toLowerCase()` is done on the frontend before sending. Meanwhile the backend does its own `trim().toLowerCase()` in `validateRepoName()`. This is fine — both sides sanitize — but worth noting the frontend validation is purely cosmetic.

## Security Findings

No security issues found.

The security-sensitive paths are well-protected:

- **CLI flag injection via `visibility`**: Validated at two layers (route handler line 98-99, `createGitHubRepo` line 72-74). Both `gh()` and `git()` use `execFile()` with array args, preventing shell metacharacter interpretation.
- **Repo name injection**: Strict whitelist regex `/^[a-z0-9][a-z0-9._-]*$/` prevents path traversal, shell metacharacters, and other injection vectors. Length capped at 100.
- **URL from GitHub API response**: The `sshUrl` returned by `gh repo create` is passed to `git remote add`. This is trusted output from GitHub's API via `gh` CLI — no user-controlled injection path.
- **No PII/secret leakage**: Error paths don't log credentials. The silent `catch` on GitHub failure deliberately avoids logging potentially sensitive error details.

## Verdict

**APPROVE** — Clean, well-tested feature with proper security hardening. The nits (silent failure UX, orphaned GitHub repo) are minor and don't block merge.

0xlny (commented):
## Review

Beau travail — le design est clean. `initProject` produit le même artefact qu'un `createProject` (bare repo + commit sur main), donc tout le flow downstream (workspace creation, sessions, git-sync, merge) est réutilisé tel quel. C'est exactement ce qu'on veut. Quelques points à adresser :

---

### 1. Repo GitHub orphelin si le push échoue (Medium)

`backend/src/projects/project-manager.ts` — `createGitHubRepo()`

Si `gh repo create` réussit mais que `git remote add` ou `git push --all` échoue ensuite, un repo vide persiste sur GitHub sans que le projet local ne le référence. Le `catch` dans `initProject` dégrade silencieusement en local-only, l'utilisateur ne sait pas qu'un repo fantôme existe sur son compte.

**Fix** — cleanup dans `createGitHubRepo` après le `gh repo create` :

```typescript
const url = stdout.trim();

try {
  await git(["remote", "add", "origin", url], bare);
  await git(["push", "--all", "origin"], bare);
} catch (err) {
  // Roll back the GitHub repo so we don't leave orphans
  await gh(["repo", "delete", name, "--yes"]).catch(() => {});
  throw err;
}

return url;
```

---

### 2. Échec silencieux de la création GitHub — pas de feedback utilisateur (Medium)

`backend/src/projects/project-manager.ts:345-347`

```typescript
} catch {
  // GitHub creation failed — project remains usable as local-only
}
```

L'utilisateur a explicitement choisi public/private, il s'attend à un repo GitHub. Si ça échoue (auth expirée, name collision, rate limit), il voit "Local only" sans aucune explication.

**Fix** — remonter un `warning` dans la réponse :

```typescript
// dans initProject, après le catch
let warning: string | undefined;
if (options.visibility) {
  try {
    const url = await createGitHubRepo(repoName, options.visibility, bare);
    state.url = url;
    await saveProject(state, dataDir);
  } catch (err) {
    warning = `GitHub repo creation failed: ${err instanceof Error ? err.message : "unknown error"}`;
  }
}
return { ...state, warning };
```

Côté frontend, afficher `warning` en toast ou bandeau jaune dans le dialog. Le projet reste utilisable, mais l'utilisateur sait pourquoi il est local-only.

---

### 3. `mkdir` après le try/catch — répertoire incomplet si ça échoue (Low)

`backend/src/projects/project-manager.ts:324-325`

```typescript
// Ces deux lignes sont APRÈS le try/catch
await mkdir(wsDir, { recursive: true });
await mkdir(logsDir, { recursive: true });
```

Si `mkdir` échoue ici (permissions, disk full), le `saveProject` s'exécute quand même et le projet est persisté avec un répertoire incomplet. Déplace ces deux `mkdir` **dans** le bloc try, juste après le `rm(tempWork)`.

---

### 4. iOS: `APIClient()` instancié sans config (Low)

`ios/HiveMobile/Views/Hub/AddProjectSheet.swift:155`

```swift
let status = try await APIClient().fetchAccountStatus()
```

Ça crée une nouvelle instance au lieu d'utiliser celle injectée via `@Environment`. Si l'utilisateur a un host/port custom configuré, cette instance pointe vers le mauvais serveur. Passe le `APIClient` en paramètre ou via l'environment comme c'est fait partout ailleurs dans l'app.

---

### Nits (non-bloquants)

- **`_init-temp` déterministe** (`project-manager.ts:296`) — si un run précédent a planté sans cleanup, le `git clone` dans ce path va échouer. Un suffixe `_init-temp-${nanoid(4)}` serait plus robuste.
- **Regex dupliquée** — `REPO_NAME_RE` est copiée entre `AddProjectDialog.tsx:30` et `project-manager.ts:48`. Un commentaire `// must match backend REPO_NAME_RE` côté frontend suffit pour éviter la dérive silencieuse.

---

Le reste est propre. La double validation de `visibility` (route + `createGitHubRepo`), la regex stricte sur le nom, le cleanup du `projectDir` en cas d'échec git, le refactor `createProjectThenWorkspace` côté frontend — tout ça est bien fait.

florian-bellotti:
### Hive: hive - PR reviewer

Let me examine the key files in detail to understand the full context before writing my review.Now let me check a couple of specific details that could affect the review.Now let me check how `createProject` in the clone path was handled before this PR (the enrichProject change), and verify the iOS APIClient pattern.Let me check one more detail about the repo name length validation on the frontend side.Good, confirmed — no length validation on either client. Now I have everything I need for the full review.

---

## Summary

PR #180 adds a "Create new repository" feature to Hive across all three platforms. The implementation is well-structured with solid defense-in-depth: repo name validation via strict regex, two-layer visibility enum whitelist (API + function), graceful degradation to local-only on GitHub failure, orphan cleanup on partial failures, and safe `execFile` array-arg CLI invocation throughout. Test coverage is thorough. The code follows existing project patterns. This PR has been hardened across 4 iterative commits and is safe to merge. The findings below are all minor/non-blocking.

## Code Review Findings

### 1. Nit: Frontend and iOS missing 100-char name length validation

**Location**: `frontend/src/components/AddProjectDialog.tsx:84-87`, `ios/HiveMobile/Views/Hub/AddProjectSheet.swift:189-192`

Both clients validate the repo name regex but don't enforce the backend's 100-character limit (`project-manager.ts:55`). A user can type a 101+ char name, the submit button enables, and the backend returns a confusing `"Repository name must be 100 characters or fewer"` error.

**Fix**: Add `name.length > 100` to the `nameError` check on frontend, and a length check in `isValidRepoName` on iOS.

### 2. Nit: `mode` field not strictly validated at the route level

**Location**: `backend/src/api/projects.ts:89`

```typescript
const mode = (body.mode as string) ?? "clone";
```

Any unknown mode value (e.g., `"delete"`, `"foo"`) silently falls through to the clone path, which will then fail with `"url is required"` — a misleading error.

**Fix**: Add an explicit check:
```typescript
if (mode !== "clone" && mode !== "create") {
  return reply.status(400).send({ error: "mode must be 'clone' or 'create'" });
}
```

### 3. Nit: iOS creates a new `APIClient()` per sheet presentation

**Location**: `ios/HiveMobile/Views/Hub/HubView.swift:69`

```swift
AddProjectSheet(api: APIClient(), ...)
```

A fresh `APIClient` is instantiated each time the sheet presents. While functional (auth reads from UserDefaults), it's inconsistent with the rest of the app which uses a shared instance from the store.

**Fix**: Pass the `ProjectStore`'s existing `api` instance, or inject it from the environment.

### 4. Note: `fetchProject` throws opaque error for local-only projects

**Location**: `backend/src/projects/project-manager.ts:205`

```typescript
if (!state.url) throw new Error("Project has no remote — nothing to fetch");
```

If any UI flow triggers a fetch on a local-only project, this surfaces as a generic error. The frontend should ideally hide/disable the fetch action when `project.url` is undefined.

### 5. Note: `enrichProject` now applied to clone-mode POST response (behavior change)

**Location**: `backend/src/api/projects.ts:110`

Previously, `POST /api/projects` (clone) returned the raw `ProjectState`. Now both modes go through `enrichProject`, adding `repoPath`, `workspacesPath`, `hasFavicon`, and `sessionCount` per workspace. This is backward-compatible (all fields are optional in the frontend type) and arguably an improvement, but it's an undocumented side effect of the refactor.

## Security Findings

No security issues found.

The PR's security posture is strong:

- **CLI injection**: Both `git()` and `gh()` use `execFile()` with array args — no shell interpolation possible regardless of input content.
- **Flag injection via `visibility`**: Mitigated by two-layer enum whitelist (API handler + `createGitHubRepo` function). Test coverage explicitly covers the `"template=evil/repo"` attack vector.
- **Repo name injection**: `REPO_NAME_RE` (`/^[a-z0-9][a-z0-9._-]*$/`) prevents flag-like patterns (no `--` prefix), shell metacharacters, and path traversal. The name is not used in path construction (paths use `proj-<nanoid>`).
- **Orphan cleanup**: GitHub repo deleted on push failure; local project dir deleted on git init failure.
- **XSS**: Warning message rendered via React JSX (auto-escaped), not `dangerouslySetInnerHTML`. iOS uses SwiftUI `Text` (also safe).
- **`writeFile` with user-controlled content**: `README.md` template uses validated repo name (alphanumeric only) — no markdown injection.

## Verdict

**APPROVE** — Well-hardened implementation with thorough test coverage and defense-in-depth security. All findings are nits/notes, none blocking.

Issue #259, opened 2026-06-21
PRD: Smooth chat resync after hub reconnect


0xlny:
## Problem Statement

When Hive regains focus after a Mac sleep/wake cycle, the desktop app now forces a hub WebSocket reconnect to recover from zombie sockets. That fixed stale streaming state, but it introduced a worse user experience: the active streaming conversation is cleared before the backend bootstrap sends the authoritative session state back.

From the user's perspective, Hive looks like it deletes the in-progress assistant response, reloads the conversation, then recreates the stream. The desired behavior is simpler: if a user starts a conversation, sees streaming begin, closes the Mac, and reopens Hive later, the conversation should look exactly like it did before sleep until the backend silently reconciles it. The UI should update without unnecessary visual changes, transitions, or empty intermediate states.

## Solution

Keep the current reconnect behavior, but make conversation resync non-destructive.

When the hub reconnects, the visible conversation state should remain unchanged. The active stream text, tool calls, agent activities, timer, input state, and tab streaming indicators derived from the active conversation should continue to look as if the stream is still running. During this internal resync window, status messages from the backend are provisional and must not change visible conversation state by themselves.

The conversation should reconcile only when it receives an authoritative event that can replace or finalize the displayed state without a glitch: persisted history, a stream snapshot, done, or cancelled. There should be no visual fallback, timeout, loading indicator, skeleton, or React timer for this resync path.

## User Stories

1. As a Hive desktop user, I want a streaming conversation to remain visible after my Mac wakes, so that I do not lose context when returning to Hive.
2. As a Hive desktop user, I want Hive to recover zombie WebSockets after sleep, so that the app starts receiving live updates again.
3. As a Hive desktop user, I want the current assistant response to stay on screen during reconnect, so that the app does not appear to delete work in progress.
4. As a Hive desktop user, I want tool calls to remain visible during reconnect, so that background activity does not visually reset.
5. As a Hive desktop user, I want the streaming timer to remain visible during reconnect, so that the UI does not change for no useful reason.
6. As a Hive desktop user, I want the chat input state to remain consistent during reconnect, so that the app does not briefly look idle while the stream is still shown.
7. As a Hive desktop user, I want the active session tab to keep its streaming indication during reconnect, so that the tab strip does not flicker.
8. As a Hive desktop user, I want Hive to apply the backend stream snapshot atomically, so that the current response updates without duplicated text or duplicated tool calls.
[...]

PR #270, opened 2026-06-30
Recover zombie hub sockets and make reconnect non-destructive

0xlny:
## Problem

Returning to the desktop app after a Mac sleep/wake (or a network change) often required a full page reload, and sometimes the conversation flashed as if it had been deleted and recreated.

Two root causes, both verified against `main`:

1. **No client-side liveness on web.** The hub WebSocket reconnects only on `ws.onclose`. A frozen TCP connection after sleep stays `readyState === OPEN` and never fires `onclose`, so the UI is stuck on stale data forever. The one recovery path (focus reconnect, #257) was reverted by #261.
2. **Destructive reconnect.** On reconnect the client wiped all in-progress streaming state (`_ws_reconnected → sessionStreams = {}` on web, `clearAllStreamingState()` on iOS), so the active assistant response visibly disappeared and was recreated. This is the behavior #259 calls out, and the reason #257 was reverted.

The `stream_snapshot` already has REPLACE semantics (since #256), so the destructive clear is vestigial.

## Change

Restore zombie-socket recovery **without** the destructive resync.

**Backend** (`types.ts`, `ws/stream.ts`)
- Add an application-level hub `ping` → `pong`. The browser WebSocket API never surfaces protocol ping/pong to JS, so the client needs an app-level probe. `HubOutgoing` becomes a union (`{workspaceId,event}` | `{type:"pong"}`).

**Web** (`lib/ws-transport.ts`, `hooks/useConversation.ts`, `main.tsx`)
- 25s heartbeat; a watchdog reconnects a socket that stays silent past the pong window (background safety net).
- `probeLiveness()` on `visibilitychange` / `focus` / `online`: actively pings and reconnects **only if the probe is unanswered within 3s**. A healthy idle conversation is never needlessly reconnected — this avoids the false positive that sank #257.
- Non-destructive resync (#259): removed the `_ws_reconnected` `sessionStreams` wipe. Active streams reconcile via the `stream_snapshot` REPLACE; the `history` reducer now drops a stale **non-streaming** stream slot (keeping only pending tool inputs), so a turn that finished during the freeze reconciles without a flash.

**iOS** (`HubStatusMonitor.swift`, `ConversationStore.swift`, `ConversationStoreCache.swift`)
- Mirror the non-destructive resync: removed `clearAllStreamingState()` on reconnect; same `history` reconcile rule. iOS keeps its existing `sendPing` heartbeat and foreground `forceReconnect` (its liveness detection was already good).

## Testing

- `npm run lint` + `npm run typecheck` clean (backend + frontend).
- Backend: **1540 tests pass** (adds a hub `ping`→`pong` test).
- Frontend transport + hook suites pass (adds heartbeat, watchdog, `probeLiveness`, and two non-destructive-resync cases). Full frontend suite: only pre-existing unrelated failure is `Sidebar.test.tsx > keeps add repository compact` (fails on baseline too).
- iOS: Swift tests added, **not compiled** (no Swift/Xcode toolchain in this environment) — needs CI.

## Notes

This is Phase 1 of a 3-phase plan. Phase 2 (light WS bootstrap / lazy tool outputs, addressing slow snapshot load — the abandoned #255 direction) and Phase 3 (backend dead-peer `pong`/terminate + removing the half-migrated history-token juggling) are follow-ups.

### DIFF
--- a/backend/src/__tests__/e2e.test.ts
+++ b/backend/src/__tests__/e2e.test.ts
@@ -59,7 +59,7 @@ async function connectHub(
     onInit: (clientWs: WebSocket) => {
       clientWs.on("message", (data: Buffer) => {
         const envelope = JSON.parse(data.toString()) as HubOutgoing;
-        if (envelope.workspaceId === workspaceId) {
+        if ("workspaceId" in envelope && envelope.workspaceId === workspaceId) {
           messages.push(envelope.event);
         }
       });
--- a/backend/src/types.ts
+++ b/backend/src/types.ts
@@ -720,10 +720,14 @@ export interface UpdateCustomAgentRequest {
 /** Client -> Server (hub-level). */
 export type HubIncoming =
   | { type: "sync_workspaces"; workspaceIds: string[] }
+  // Application-level liveness probe. The browser WebSocket API never exposes
+  // protocol-level ping/pong to JS, so clients send this to actively detect a
+  // frozen-but-OPEN socket (e.g. after the OS wakes from sleep).
+  | { type: "ping" }
   | { workspaceId: string; event: WsIncoming };
 
-/** Server -> Client (hub-level). Every outgoing event is tagged with its workspace. */
-export interface HubOutgoing {
-  workspaceId: string;
-  event: WsOutgoing;
-}
+/** Server -> Client (hub-level). Workspace events are tagged with their workspace. */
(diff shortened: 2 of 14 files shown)


0xlny:
Folded into #272 (single PR for the whole streaming overhaul, backend + web + iOS).

PR #281, opened 2026-07-01
Optimize iOS data & battery usage across backend, iOS, and web

florian-bellotti:
## Summary

Reduces cellular-data and battery consumption on the iOS client — previously heavy when switching workspaces/projects/discussions and while streaming. Spans backend, iOS, and web; all changes are backward-compatible across surfaces (web/older clients that don't send the new WS fields keep the prior behavior).

## Backend
- **Compression**: brotli/gzip on responses (threshold 1KB).
- **ETag + 304**: conditional GET for JSON responses; the hook now respects a route's own `Cache-Control`, and the secret-bearing `/env` route sends `no-store`.
- **Scoped WS subscription (C13)**: high-frequency stream events (`text_delta`/`thinking`/`tool_use`/`tool_result`/`agent_activity`/`stream_snapshot`) are only delivered to hub sockets that focus the workspace; low-frequency events (status/done/branch/diff/pr) still broadcast to all. Undefined focus = receive everything.
- **Event-driven `pr_status` push (C17)**: folded into the git-sync loop, gated to workspaces with a live hub subscriber, 60s TTL, flap-guarded against transient `gh` errors. The `gh` PR fetch now has an **8s timeout** so a stuck call can't freeze the shared sync loop.
- `?since` message cursor, `sharp` image resize (`?w=`), versioned favicon caching.

## iOS
- **Connection**: probe-liveness on refresh instead of always reconnecting; reconnect idempotency; backoff cap + jitter; foreground-refresh threshold; `performConnect` cancels any armed reconnect task.
- **Rendering/state**: `@Observable` stream state so text deltas don't recompute task/goal/agent derivations; derivations recomputed at the right lifecycle points; LRU session cache + memory-warning eviction; plain-text streaming bubble; `NSCache` for tool-input JSON.
- **Networking**: configured shared `URLSession` with `URLCache`; ImageIO downsampling; incremental history via `?since`; freshness-gated history refetch (invalidated on background-session updates).
- **PR status**: consume the `pr_status` push; the bulk PR poll is now a fallback for non-subscribed workspaces only (WS is authoritative), removing a dual-source clobber race.

## Web
- Type alignment for the new `pr_status` / `focusWorkspaces` protocol fields; consume `pr_status` push into the React Query cache. (C13 data savings are iOS-scoped by design; web behavior is unchanged.)

## Verification
- **Backend**: typecheck + lint clean; full suite **1555 passing** (incl. new C13 focus-filter and C17 gating tests, and a `no-store` security assertion).
- **iOS**: `swift test` **97 passing**; full `xcodebuild` (iPhone 17 sim) succeeds.
- **Web**: typecheck + lint clean; **1470 passing**.
- **Live e2e**: compression (`content-encoding: br`/`gzip`), ETag `304`, and WS `101` handshake verified over the wire and in-browser (no console errors).

## Notes
- Two multi-agent review passes were run: a simplicity review (applied the surviving simplifications) and a bug/perf audit that caught and fixed 8 issues — headline being the `gh`-without-timeout freeze risk above.
- Excludes local-only working-tree changes (`.gitignore`, Xcode `DEVELOPMENT_TEAM` signing, `.mcp.json`).


florian-bellotti:
Follow-up commit `070f76c` from a third multi-agent re-audit (bug + perf + simplicity, weighted to the previously-applied fixes):

- **Fixed a C13 regression** the earlier passes missed: an already-subscribed but *unfocused* workspace that streams has its high-frequency deltas filtered; a later focus switch only reassigned `focusWorkspaces` without replaying the snapshot, so the in-progress stream prefix was missing until the turn completed. `handleSyncWorkspaces` now replays the streaming snapshot for already-subscribed workspaces that (re)enter focus — idempotent, streaming-only, no-op otherwise (+1 injectWS test).
- Corrected a stale comment (`SessionStreamState` value-type claim after the struct→class conversion).

The audit found **no perf regressions and no over-engineering** in the prior fixes. Verified: backend **1556** ✓, iOS `swift test` **97** ✓, typecheck/lint clean.

0xlny:
Follow-up audit fixes applied in `050448a`.

Findings addressed:

1. **Bootstrap stream snapshots bypassed focus filtering**
   - Problem: live broadcasts honored `focusWorkspaces`, but direct bootstrap sends still delivered high-frequency `stream_snapshot` payloads for unfocused workspaces.
   - Fix: `sendToHub()` now applies the same `hubReceivesEvent()` filter as broadcast paths, and the C13 tests now assert that unfocused bootstraps receive low-frequency `status` but no `stream_snapshot` until focus changes.

2. **Foreground/pull refresh no longer replayed hub bootstrap on healthy sockets**
   - Problem: replacing reconnect with a liveness probe saved battery, but a successful ping did not replay status/snapshot/branch/diff/PR/script bootstrap state, so iOS could remain stale after backgrounding.
   - Fix: added optional `forceBootstrap` to `sync_workspaces`. iOS now probes the existing socket and, when healthy, requests a full bootstrap over that socket instead of reconnecting. Backend tests cover forced bootstrap on an already subscribed workspace and duplicate snapshot prevention.

3. **PR status cache survived workspace branch changes**
   - Problem: branch changes cleared only the PR fetch TTL, leaving the previous branch's cached PR payload available for bootstrap, especially after a later `gh` error.
   - Fix: branch changes now clear `prFetchedAt`, `prStatusCache`, and `latestPrStatus`. Added a regression test for old-PR removal after branch rename plus a failed PR fetch.

Verification run locally:

- `git diff --check`
- `npm run test --workspace @hive/backend -- src/ws/stream.test.ts`
- `npm run test --workspace @hive/backend -- src/services/git-sync.test.ts`
- `npm run lint --workspace @hive/backend`
- `npm run typecheck --workspace @hive/backend`

Note: iOS Swift/Xcode checks could not be run in the current environment because `swift` and `xcodebuild` are unavailable here. The Swift protocol/store changes were reviewed by inspection and should still be validated by CI/macOS.

PR #299, opened 2026-07-02
Optimize iOS streaming render performance (#283)

### LINKED ISSUE #283: iOS streaming render performance: token coalescing, Equatable models, derivation caching
## What to build

Make agent streaming cheap on iOS: coalesce incoming tokens in the conversation store on a ~50ms cadence, make chat models Equatable so unchanged bubbles skip re-rendering, cache and guard all conversation derivations, throttle Hub ambient work, and eliminate the empty-history crash. Behavior-preserving; users only see smoother streaming and better battery.

## Acceptance criteria

- [ ] Text/thinking deltas are buffered in the conversation store and flushed to observable state on a ~50ms cadence with an injectable flush control for tests; the view-side throttle is deleted
- [ ] Chat message, tool call, agent activity, and image attachment models are Equatable; unchanged bubbles are not re-rendered during streaming
- [ ] Derivations (tasks, background agents, sub-agent status) recompute only for the active session and exactly once after history application; a shared parsed-tool-input cache replaces repeated JSON parsing; the activity-to-tool-call conversion is cached
- [ ] Dismissed-tool-call ids are derived in the store with an empty-history guard (regression test for the 0..<(-1) trap)
- [ ] Hub activity marking ignores token-delta-class events; the visible-PR-workspaces key is memoized; message timestamp formatters are static; activity indicators are capped at 30 Hz without the per-frame shadow
- [ ] `cd ios && swift test` passes with the new store tests

## Blocked by

None - can start immediately

---

## Problem Statement

When an agent is streaming a response, the Hive iOS app becomes visibly sluggish. Scrolling the chat stutters, the keyboard and buttons respond late, animations hitch, and the device gets warm and drains battery faster than it should. The slowdown gets worse the longer a conversation is: sessions with many messages and heavy tool activity (large diffs, many tool calls) can make the chat nearly unresponsive while tokens arrive. The Hub screen, which sits under the chat in the navigation stack, also does redundant work during streaming even though it is not visible, compounding the load. In one edge case the app can crash outright: if a new turn starts from another client before the chat's history has loaded, an internal computation over an empty message list traps.

Users experience all of this as "the app is slow while the agent is working" — exactly the moment they most want to watch output arrive smoothly.

## Solution

Make streaming cheap by ensuring each incoming token updates only what actually changed, at a bounded rate:

- Incoming text and thinking tokens are coalesced inside the conversation store and applied to the UI on a fixed ~50ms cadence, so the screen refreshes at most ~20 times per second regardless of token rate.
- Message rows become skippable: SwiftUI can compare a row's content and re-render only the bubble that changed, instead of re-rendering, re-parsing Markdown for, and re-merging tool calls into every visible bubble on every token.
- Expensive derived data (task lists, background-agent status, sub-agent status, tool-call synthesis from Codex activities, dismissed tool-call ids) is computed once per relevant change, cached, and only recomputed for the session the user is actually viewing.
- Ambient work is throttled: the Hub stops re-sorting workspaces at token rate, timestamp formatting reuses cached formatters, and animated activity indicators redraw at 30 Hz without per-frame shadow effects.
- The empty-history crash path is eliminated with an explicit guard.

None of this changes what the user sees — only how smoothly and efficiently it renders.

## User Stories

1. As a Hive mobile user, I want the chat to scroll smoothly while an agent is streaming, so that I can read earlier messages without stutter.
2. As a Hive mobile user, I want streamed text to appear at a steady, readable cadence, so that the output feels fluid instead of janky.
3. As a Hive mobile user, I want the keyboard and send button to stay responsive during a long agent turn, so that I can queue my next message without waiting.
4. As a Hive mobile user, I want long tool-heavy sessions to perform as well as short ones, so that conversation length does not degrade the app.
5. As a Hive mobile user, I want my phone to stay cool and my battery to last during extended agent sessions, so that monitoring work does not drain my device.
6. As a developer monitoring agents, I want the app not to crash when another client starts a turn in a session I just opened, so that multi-client workflows are reliable.
7. As a developer monitoring agents, I want Codex sessions with large file diffs to render without per-token slowdowns, so that reviewing agent edits stays practical on mobile.
8. As a developer monitoring agents, I want task and background-agent status to stay accurate without lag, so that derived views reflect the conversation promptly.
9. As a Hive mobile user, I want the Hub screen to stay quiet while I am in a chat, so that off-screen views do not steal CPU from the conversation I am watching.
10. As a developer monitoring agents, I want events for background (non-active) sessions to be recorded cheaply without triggering full recomputation, so that parallel agent sessions do not multiply main-thread work.
11. As a Hive mobile user, I want activity indicators to animate efficiently, so that a simple spinner does not consume disproportionate GPU and battery.
12. As a Hive mobile user, I want message timestamps to render without measurable cost, so that scrolling through history stays fast.
13. As a Hive mobile user, I want unchanged message bubbles to be left alone when new tokens arrive, so that only the streaming bubble updates.
14. As a developer monitoring agents, I want derivation work to run exactly once after history loads, so that opening a long session does not pay the same cost twice.
15. As a Hive mobile user, I want the app to feel equally responsive on older devices, so that streaming performance does not require the latest hardware.
16. As a developer maintaining the iOS client, I want token coalescing to live in the store rather than the view layer, so that the throttle cannot be bypassed by view observation patterns.
17. As a developer maintaining the iOS client, I want the coalescing cadence to be controllable in tests, so that streaming behavior is deterministic and regressions are caught in CI.
18. As a Hive mobile user, I want the Hub's pull-request section to update only when its underlying data changes, so that returning to the Hub after a chat is instant.

## Implementation Decisions

- **Store-side token coalescing.** Text and thinking deltas received from the WebSocket hub are appended to an internal buffer in the conversation store; the buffer is flushed into observable state on a ~50ms cadence. The existing view-side 50ms throttle (driven by view `onChange` of raw per-token properties) is deleted — it is ineffective because the view body observes the raw properties directly and re-runs per frame anyway. After coalescing, the chat body re-runs at most ~20 times per second during streaming. The flush mechanism is injectable (configurable interval or manual flush hook) so tests can drive it deterministically; production uses the default cadence.
- **Equatable models.** Chat message, tool call, agent activity, and image attachment models gain `Equatable` conformance so SwiftUI's row diffing can skip unchanged rows. Row views compare by value; only the actively streaming bubble re-renders per flush. This removes per-token Markdown re-parsing and tool-call re-merging for stable bubbles.
- **Cached activity-to-tool-call conversion.** The synthesis of tool calls from Codex agent activities (which JSON-encodes full file diffs) is computed once per activity change and cached in the store, not recomputed on every property access per bubble per render.
- **Guarded, deduplicated derivations.** Task, background-agent, and sub-agent derivations recompute only for the active session; events for non-active sessions update stored state without triggering a full conversation rescan. The duplicate recompute after history application (property `didSet` plus explicit call) is collapsed into a single trigger. All derivations share one parsed-tool-input cache so tool JSON is parsed at most once per tool call per content version.
- **Store-derived dismissed tool calls.** The dismissed-tool-call id set moves from an O(n) per-body-pass computation in the view into the store, recomputed only when messages change, with an explicit empty-history guard that eliminates the invalid-range crash.
- **Hub activity throttling.** Hub activity tracking no longer writes its observed dictionary for token-delta-class frames; alternatively (implementer's choice) it writes only when the stored timestamp is older than ~5 seconds. Either way, the Hub stops re-sorting workspaces at token rate while alive under the chat.
- **Memoized visible-PR-workspaces key.** The Hub's visible-PR-workspaces key (currently rebuilt via section iteration, sort, and join on every body pass) is memoized and recomputed only when its inputs change.
- **Static formatters and 30 Hz timelines.** Message timestamp formatting uses cached static date formatters, matching the pattern used elsewhere in the app. Activity indicators use a minimum `TimelineView` interval of 1/30s and drop the per-frame shadow blur.
- All changes are behavior-preserving from the user's perspective; the store remains the single owner of conversation state and views stay thin.

## Testing Decisions

- All testable logic sits at the boundary "WS event in → observable store state out" (conversation store and hub status monitor), following the existing Swift Testing suites for the conversation store, task derivation, and background-agent derivation.
- The one new approved seam is the injectable streaming-buffer flush control: tests feed a sequence of token-delta events, then trigger a manual flush (or advance the injected interval) and assert the resulting observable message content. This proves coalescing is deterministic without timing-based flakiness.
- Tests to add:
  - Token deltas do not mutate observable message state until a flush; a flush applies the full buffered text in order; interleaved text and thinking deltas flush correctly.
  - `Equatable` conformance: two models with identical content compare equal; a single-field change compares unequal (guards against accidental identity-only equality).
  - Derivations do not recompute for events targeting a non-active session, and recompute exactly once after history application.
  - The cached activity-to-tool-call conversion returns a stable value across repeated accesses and updates when the underlying activity changes.
  - Dismissed-tool-call derivation on an empty history returns an empty set (regression test for the crash) and behaves correctly when a turn begins before history arrives.
  - Hub activity marking ignores token-delta-class events (or respects the ~5s staleness window), asserted via hub monitor state.
- Assertions target external, observable state after events — never internal buffers, call counts on private members, or view internals.
- Purely visual items (30 Hz indicator cadence, shadow removal, static formatter output) are verified manually; views stay thin so nothing there needs unit coverage.

## Out of Scope

- Network-layer changes: incremental history fetch and off-main WebSocket decoding are covered by the separate "iOS network performance" PRD.
- Reduce Motion / accessibility motion support (separate PRD).
- Any visual or UX change: layouts, copy, animation designs, and interaction behavior are unchanged except for the imperceptible cadence caps described above.
- Backend or protocol changes; web frontend changes.

## Further Notes

- The ~50ms flush cadence and the ~5s hub-activity staleness window are tuning defaults, not contracts; they may be adjusted later without revisiting this PRD, provided the injectable test seam keeps them controllable.
- The shared parsed-tool-input cache introduced here is also touched by the "iOS network performance" PRD (cache-key correctness for growing synthetic tool calls); coordinate to avoid conflicting cache designs if both land close together.
- Finding 9 (memoized PR-workspaces key) is newer than the rest of the audit; confirm it still reproduces on main at implementation time before building the memoization.
- Success criteria for manual verification: during a fast streaming session on a physical device, chat scrolling stays smooth, Instruments shows body re-evaluation bounded near the flush cadence rather than the token rate, and the Hub shows no per-token work while a chat is on top.

florian-bellotti:
Closes #283

## Summary

Streaming a long turn on iOS previously mutated observable state and re-ran expensive derivations on every WS token, re-parsed tool-input JSON on every render pass, and drove uncapped animation timelines. This PR moves token coalescing into the store, makes hot models Equatable so SwiftUI can skip unchanged bubbles, caches derivations and JSON parses, and caps hub/indicator work — with no visual or UX changes beyond the cadence caps.

- **Store-side 50ms token coalescing with injectable flush; view throttle deleted** — `ConversationStore` buffers `text_delta`/`thinking` fragments in `@ObservationIgnored` per-session buffers and applies them on a 50ms cadence (`init(streamFlushInterval:)`, `nil` = manual flush for tests). Buffers flush before `done`/`cancelled`/`status`/`tool_input_required` and are discarded on `stream_snapshot` (authoritative) and `user_message`. ChatView's render-throttle state machine is deleted; the view reads `store.currentText`/`store.currentThinking` directly.
- **`ChatMessage` and `ToolCall` are `Equatable`** — synthesized `==` (all stored fields are value types), letting SwiftUI diff message bubbles cheaply.
- **Derivation gating, single post-history recompute, and shared caches** — task/goal/background-agent derivations now run only for the active session; applying fetched history recomputes exactly once (stream-slot reconciliation moved before the `messages` assignment). Tool-input JSON parsing goes through one shared content-keyed `NSCache` (`ToolInputParsing.swift`) used by `TaskDerivation`, `SubAgentStatus`, and `MessageBubble`; the `AgentActivity` → `[ToolCall]` conversion is cached by activity value with early-return for kinds that never produce tool calls.
- **Store-derived dismissed AskUserQuestion ids with empty-history crash fix** — `deriveDismissedToolCallIds(from:)` uses `zip(messages, messages.dropFirst())` (inherently empty-safe), replacing ChatView's `0..<(msgs.count - 1)` loop that trapped on empty history when a turn started from another client before history loaded.
- **Hub and UI hot paths** — new pure `hubActivityMarking(for:)` helper classifies WS events for last-activity marking and ignores per-token deltas; `HubView` computes sections and visible PR-workspace ids once per body pass and uses the `[String]` directly as the `.task` id (sort+join key deleted); `MessageBubble` timestamp formatters are static; `AgentActivityIndicator` and the chat pulse `TimelineView`s are capped at 30fps and the indicator's per-frame shadow is removed.
- **Tests** — 5 new suites (streaming buffer, model equatability, derivation gating/dismissed ids, activity tool-call cache, hub activity marking); existing delta-sending store tests switched to manual flushing so they never rely on the async auto-flush.

## Behavior notes

- **Deliberate, spec-mandated change:** hub `lastActivityAt` no longer advances on `text_delta`/`thinking` events, so Hub sort order updates at tool-use/turn boundaries during pure-text streams instead of on every token. Sorting-only impact and self-correcting (the next tool use, status, or turn completion marks activity).
- **Fixed latent bug:** the old MessageBubble parse cache was keyed by tool id only, so activity-synthesized tool calls whose input grows across upserts (stable id, growing diff) could render stale input. The shared cache is content-keyed, so updated input is always re-parsed.

## Testing

- `cd ios && swift test` → **113 tests / 18 suites green** (baseline on main: 94/13; 19 new tests across 5 suites).
- `xcodebuild build` for the iOS Simulator → **BUILD SUCCEEDED** (covers the view files SwiftPM excludes from the test target).
- Both checks were also re-run independently by two reviewers (conformance + adversarial), with zero blockers/majors.

### DIFF
--- a/ios/HiveMobile/Models/AgentActivity.swift
+++ b/ios/HiveMobile/Models/AgentActivity.swift
@@ -260,6 +260,7 @@ private final class CachedActivityToolCalls {
 private let activityToolCallsCache: NSCache<NSString, CachedActivityToolCalls> = {
     let cache = NSCache<NSString, CachedActivityToolCalls>()
     cache.countLimit = 512
+    cache.totalCostLimit = 2_000_000
     return cache
 }()
 
@@ -279,7 +280,8 @@ extension AgentActivity {
         let computed = computeToolCalls()
         activityToolCallsCache.setObject(
             CachedActivityToolCalls(activity: self, toolCalls: computed),
-            forKey: key
+            forKey: key,
+            cost: activityToolCallsCacheCost(computed)
         )
         return computed
     }
@@ -322,6 +324,17 @@ extension AgentActivity {
         case .planUpdate, .goalUpdate, .imageView, .imageGeneration, .diagnostic, .unknown:
             return []
         }
+    }
+}
+
+private func activityToolCallsCacheCost(_ toolCalls: [ToolCall]) -> Int {
+    toolCalls.reduce(0) { total, tool in
+        total
+            + tool.id.utf16.count
+            + tool.name.utf16.count
+            + tool.input.utf16.count
+            + (tool.output?.utf16.count ?? 0)
+            + (tool.parentToolUseId?.utf16.count ?? 0)
     }
 }
 
--- a/ios/HiveMobile/Models/AgentActivity.swift
+++ b/ios/HiveMobile/Models/AgentActivity.swift
@@ -247,8 +247,44 @@ enum AgentActivity: Codable, Equatable, Identifiable {
     }
 }
 
+private final class CachedActivityToolCalls {
+    let activity: AgentActivity
+    let toolCalls: [ToolCall]
+
+    init(activity: AgentActivity, toolCalls: [ToolCall]) {
+        self.activity = activity
+        self.toolCalls = toolCalls
+    }
+}
+
+private let activityToolCallsCache: NSCache<NSString, CachedActivityToolCalls> = {
+    let cache = NSCache<NSString, CachedActivityToolCalls>()
+    cache.countLimit = 512
+    return cache
+}()
+
 extension AgentActivity {
     var toolCalls: [ToolCall] {
+        switch self {
+        case .planUpdate, .goalUpdate, .imageView, .imageGeneration, .diagnostic, .unknown:
+            return []
+        case .commandExecution, .fileChange:
+            break
+        }
+
+        let key = id as NSString
+        if let cached = activityToolCallsCache.object(forKey: key), cached.activity == self {
+            return cached.toolCalls
+        }
+        let computed = computeToolCalls()
+        activityToolCallsCache.setObject(
+            CachedActivityToolCalls(activity: self, toolCalls: computed),
+            forKey: key
+        )
+        return computed
+    }
+
+    private func computeToolCalls() -> [ToolCall] {
         switch self {
         case .commandExecution(let activity):
             return [toolCall(for: activity)]
--- a/ios/HiveMobile/Models/Models.swift
+++ b/ios/HiveMobile/Models/Models.swift
@@ -219,7 +219,7 @@ struct FileMention: Codable, Equatable {
     let relativePath: String
 }
 
-struct ToolCall: Codable, Identifiable {
+struct ToolCall: Codable, Equatable, Identifiable {
     let id: String
     let name: String
     let input: String
@@ -256,7 +256,7 @@ struct ToolCall: Codable, Identifiable {
     }
 }
 
-struct ChatMessage: Codable, Identifiable {
+struct ChatMessage: Codable, Equatable, Identifiable {
     let id: String
     let sessionId: String
     let role: MessageRole
--- a/ios/HiveMobile/Models/WebSocketTypes.swift
+++ b/ios/HiveMobile/Models/WebSocketTypes.swift
@@ -324,6 +324,33 @@ enum WsOutgoing: Decodable {
     }
 }
 
+// MARK: - Hub Activity Marking
+
+/// How a hub-level event should update a workspace's last-activity timestamp
+/// (used for hub sorting). Per-token events are ignored so streaming does not
+/// write observable hub state on every fragment.
+enum HubActivityMarking: Equatable {
+    case ignore
+    case markNow
+    case markLatestMessageTimestamp
+    case markIfStreaming
+}
+
+func hubActivityMarking(for event: WsOutgoing) -> HubActivityMarking {
+    switch event {
+    case .textDelta, .thinking,
+         .branchInfo, .diffStats, .prStatus, .scriptStatus, .planModeChanged, .streamSnapshot:
+        return .ignore
+    case .history:
+        return .markLatestMessageTimestamp
+    case .status:
+        return .markIfStreaming
+    case .toolUse, .toolResult, .agentActivity, .toolInputRequired, .toolInputResolved,
+         .done, .error, .cancelled, .userMessage, .unknown:
+        return .markNow
+    }
+}
+
 // MARK: - AnyCodableValue (for decoding unknown JSON)
 
 enum AnyCodableValue: Codable {
--- a/ios/HiveMobile/Stores/ConversationStore.swift
+++ b/ios/HiveMobile/Stores/ConversationStore.swift
@@ -107,7 +107,7 @@ final class ConversationStore {
     private(set) var backgroundAgents: BackgroundAgentsState =
         deriveBackgroundAgents(from: [], activeToolCalls: [])
     private(set) var dismissedToolCallIds: Set<String> = []
-    private(set) var derivationRunCount = 0
+    @ObservationIgnored private(set) var derivationRunCount = 0
 
     private func recomputeDerivations() {
         derivationRunCount += 1
--- a/ios/HiveMobile/Stores/ConversationStore.swift
+++ b/ios/HiveMobile/Stores/ConversationStore.swift
@@ -70,6 +70,16 @@ final class ConversationStore {
     /// Provider locked for this session (pushed via WS status events).
     var lockedProvider: String?
 
+    private let streamFlushInterval: TimeInterval?
+    @ObservationIgnored private var pendingTextBySession: [String: String] = [:]
+    @ObservationIgnored private var pendingThinkingBySession: [String: String] = [:]
+    @ObservationIgnored private var flushTask: Task<Void, Never>?
+
+    /// `nil` disables the scheduled auto-flush (tests flush manually).
+    init(streamFlushInterval: TimeInterval? = 0.05) {
+        self.streamFlushInterval = streamFlushInterval
+    }
+
     // MARK: - Computed
 
     /// The active session's stream state (convenience accessor).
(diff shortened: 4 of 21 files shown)


florian-bellotti:
## E2E validation — final results (rebased head d34c793)

Full re-run of all six scenarios against the current head (d34c793, which integrates the #282 List-based transcript), on an iPhone 17 Pro simulator, throwaway local backend with isolated data dir, provider claude:haiku-4-5. App process pid stayed constant across the entire run (~10 turns, repeated Hub↔chat navigation) — no crash or restart.

1. **Build & launch smoke** — PASS. Incremental rebuild from d34c793, 0 warnings, app reaches the Hub.
2. **Connect & navigate** — PASS. Local backend via Settings > Connection.
3. **Streaming + mid-stream scroll on the List transcript** — PASS. Auto-follow during streaming; scrolling back through a long transcript during an active turn is glitch-free; final rendered 4086-char reply matches REST history byte-for-byte including the tail — no text lost to the 50 ms coalescing.
4. **Crash regression** — PASS. Turns started from another client (hub WS) streamed live into a freshly opened session and into an open ChatView, repeatedly, with zero crashes. The exact status-beats-REST race is not externally forceable; the unit regression test covers that window.
5. **Multi-client second turn** — PASS. In-app message after out-of-band turns streamed cleanly with intact multi-turn transcript.
6. **Hub during streaming** — PASS. Animated activity indicator on the streaming workspace, smooth mid-stream navigation, no per-token Hub work as specced.

No visual glitches, dropped tails, or stutter observed. E2E verdict: **GO** on the rebased branch — both the #282 List transcript and the #283 render changes behave correctly together.

PR #301, opened 2026-07-06
Add iOS message copy actions (#291)

### LINKED ISSUE #291: iOS copy actions for chat messages
## What to build

Platform-standard message copying: a long-press context menu with Copy on both user and assistant messages, selectable user message text, and a comfortably enlarged hit area on the existing assistant copy button (visuals unchanged). Copy always yields plain text.

## Acceptance criteria

- [ ] Long-press on any message (user or assistant) shows a context menu with Copy
- [ ] User message text is selectable
- [ ] The assistant copy button's tappable area is enlarged toward the 44pt guideline without visual change; both affordances produce the same clipboard result
- [ ] Multi-paragraph copies preserve line breaks
- [ ] `cd ios && swift test` passes (plain-text derivation covered if store/model logic is introduced)

## Blocked by

None - can start immediately

---

## Problem Statement

Copying content out of a Hive conversation on iOS is inconsistent and physically difficult. Assistant messages offer a small copy button in the message footer, but its tappable area is 14×14 points — far below the 44-point minimum Apple recommends — so tapping it reliably takes several attempts. User messages have no copy affordance at all and their text cannot even be selected, yet reusing a previous prompt is one of the most common workflows in an agent app: users iterate on prompts, resend variations, and move prompts between conversations. There is also no long-press context menu on any message, which is the pattern iOS users instinctively try first when they want to copy something.

## Solution

Bring message copying up to platform conventions. Every message — user and assistant — gains a long-press context menu with a Copy action, matching what users already do in Messages and every mainstream chat app. User message text becomes selectable, matching the selection behavior assistant messages already have. The existing copy button on assistant messages stays visually identical but its tappable area is enlarged to a comfortable target, so users who prefer the button stop fighting it. Copy always places the message's text content on the clipboard as plain text.

## User Stories

1. As a user, I want to long-press any message and see a Copy action, so that I can copy content using the gesture I already know from other iOS apps.
2. As a user iterating on a prompt, I want to copy a previous user message, so that I can paste it into the composer, edit it, and send a refined version.
3. As a user, I want to copy one of my prompts from one conversation and reuse it in another workspace or session, so that good prompts are portable.
4. As a user, I want to select a portion of a user message's text, so that I can extract just the relevant fragment instead of the whole message.
5. As a user, I want the copy button on assistant messages to respond on the first tap, so that copying agent output is not an exercise in precision.
6. As a user, I want the copy button to look exactly as it does today, so that the interface does not get visually heavier to fix a touch-target problem.
7. As a user copying an assistant message, I want the copied text to be the message's content as plain text, so that pasting into the composer or another app gives clean, usable text.
8. As a user, I want the context menu Copy on an assistant message and the footer copy button to produce the same clipboard result, so that the two affordances are interchangeable.
9. As a user long-pressing a message, I want the standard context-menu preview and feel, so that the interaction is familiar and predictable.
10. As a user with larger fingers or motor-accessibility needs, I want all copy affordances to meet comfortable touch-target sizes, so that the app is usable without frustration.
11. As a user, I want copying a multi-paragraph message to preserve its line breaks, so that structure survives the round trip through the clipboard.

## Implementation Decisions

- A long-press context menu with a single Copy action is added to both user and assistant message bubbles. This is the standard iOS context-menu pattern, not a custom overlay.
- Copy places the message's text content on the system clipboard as plain text. For assistant messages rendered as Markdown, the copied content is the message's underlying text, consistent with what the existing copy button produces today.
- User message text becomes selectable, mirroring the selection behavior assistant message text already has.
- The existing assistant-message copy button keeps its current visual size and appearance; its hit area is enlarged via padding and content-shape expansion to a comfortable target in line with the 44-point guideline. Neighboring footer elements must not overlap the enlarged hit area.
- The context menu contains only Copy for now; the menu is deliberately introduced as the extensible home for future message actions.
- No haptic or toast confirmation is added beyond what the existing copy button already does; behavior parity between the two affordances is the rule.

## Testing Decisions

- Store-testable parts are minimal for this PRD: if any logic is introduced for deriving a message's copyable plain text (for example, from an assistant message's content blocks), that derivation lives at the store/model layer and gets Swift Testing cases asserting input message to expected clipboard string, including multi-paragraph and code-fenced content. These follow the prior art of the existing store test suites and assert output text, not rendering internals.
- The interaction surface — long-press menu presentation on both message types, text selection on user messages, the enlarged tap target on the copy button, and actual clipboard contents after each affordance — is view-layer and is verified manually on device, including a check that long-press on links or selectable text inside bubbles does not conflict with the context menu.

## Out of Scope

- Additional context-menu actions such as edit-and-resend or quote-reply — noted as future work; the menu introduced here is their eventual home.
- Copy affordances for individual tool-call rows or other non-message conversation items.
- Copy-as-Markdown versus copy-as-rendered-text options; a single plain-text copy is the behavior.
- Any changes to the web client's copy behavior.

## Further Notes

The 14×14-point hit area on the existing copy button is the most acute usability defect here and is worth fixing even independently of the context menu. The context menu is nonetheless the primary solution because it is discoverable through instinct rather than instruction, and it establishes a pattern the rejected-for-now actions (edit-and-resend, quote-reply) can slot into later without another interaction redesign.

florian-bellotti:
Closes #291

## Summary

- Assistant messages: completed messages render through a new `SelectableMarkdownText` UIKit representable (`UITextView`, selectable, non-editable, self-sizing) giving true Apple-native text selection — long-press selects at the touch location with loupe, draggable handles, and the system edit menu. No context menu on assistant messages.
- Streaming assistant message keeps the existing MarkdownUI rendering path on purpose, preserving the PR #283 streaming render optimizations; the swap to the selectable view happens at stream completion with matched fonts/colors.
- User messages (bubble): iMessage-style lift/zoom on long-press via `.contextMenu` with a single Copy item (doc.on.doc); `.contentShape(.contextMenuPreview, ...)` clips the lift preview to the rounded bubble shape. No inline text selection on user bubbles.
- Shared `ChatMessage.clipboardText` derivation (verbatim `content` passthrough) used by the bubble Copy action and the footer copy button, covered by 4 unit tests.
- Footer copy button hit area enlarged to 44x38 via `.padding(.horizontal, 15).padding(.vertical, 12)` + `.contentShape(Rectangle())`; icon styling unchanged.

## Accepted trade-offs

- The `NSAttributedString` markdown converter is chat-grade: code blocks and tables render simplified (styled monospaced text, no horizontal scrolling or table grid) compared to the MarkdownUI theme.
- Native selection copies the rendered text (markdown syntax stripped), while the footer copy button copies the raw markdown via `clipboardText`.

## Test results

- `cd ios && swift test`: 121 tests in 19 suites passing, including 4 `MessageClipboardTextTests` (single paragraph, multi-paragraph line breaks, code fences, empty content).
- `xcodebuild -project HiveMobile.xcodeproj -scheme HiveMobile -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO build`: BUILD SUCCEEDED (compiles the views including the new representable).

Note: `swift test` does not compile SwiftUI views, hence the xcodebuild check and the manual checklist below.

## Manual verification checklist (on device)

- [ ] Assistant message: long-press starts native selection at the touch location with loupe and draggable handles; the system edit menu offers Copy/Select All; no custom context menu appears.
- [ ] Assistant message rendering after stream completion visually matches the streaming MarkdownUI rendering closely (body text, headings, bold/italic, inline code, code blocks, lists, links, blockquotes); the swap is not jarring.
- [ ] User bubble: long-press lifts the bubble iMessage-style, the preview is clipped to the rounded bubble shape, and the Copy item puts the exact message text on the clipboard; no inline selection triggers.
- [ ] Footer copy button responds on first tap, enlarged hit area does not overlap the timestamp/duration, and it copies the raw markdown with line breaks preserved.
- [ ] Links inside assistant messages open on tap and are tinted with the app accent.


florian-bellotti:
Pushed `482c5e9` — fixes a nested-list rendering bug in the new `SelectableMarkdownText`: nested list items lost their bullet/number prefix entirely (`1. First / - alpha / - beta` rendered alpha/beta as bare indented text). Root cause: `PresentationIntent.components` is ordered innermost→outermost, and the `listItem` case kept the last (outermost) match while `isOrdered` kept the first (innermost) — so all nested siblings inherited the outer item's identity and the prefix-dedup skipped them.

Changes:
- `listItem` now keeps the first (innermost) match, mirroring the existing `nearestListIsOrdered` guard.
- Block-context logic moved to `ios/HiveMobile/Stores/MarkdownStructure.swift` (registered in `Package.swift`) so `swift test` compiles it — it was previously untestable in the excluded Views folder.
- 5 new tests in `MarkdownStructureTests` covering nested/flat/non-list cases; verified they fail against the old logic.

`swift test`: 126/126 pass. `xcodebuild` (iOS Simulator): BUILD SUCCEEDED.

florian-bellotti:
Follow-up commit 612cc29: fixes the user bubble jumping left while the copy context menu dismisses.

**Root cause**: SwiftUI's `.contextMenu` computes a wrong dismissal target inside List cells — the dismiss preview platter gets re-centered horizontally in the row instead of anchored to the trailing-aligned bubble (measured: exactly `(rowWidth - bubbleWidth) / 2`). The platter sits at that wrong position until the interaction ends (~1-2s), then the real bubble reappears in place — hence the visible jump. No SwiftUI configuration fixes it (both `contentShape` placements were tested and screenshot-verified).

**Fix**: the bubble menu now uses `UIContextMenuInteraction` directly (the same native engine `.contextMenu` wraps) and supplies both targeted previews via the official delegate hooks, anchored to the bubble's real position. Menu, blur, haptics and animations are all still system-rendered.

Verified in the iOS 26.5 simulator with automated long-press/dismiss + screenshot bursts: highlight identical to before, dismiss pixel-stable, Copy action confirmed via pasteboard.

Known trade-off: VoiceOver exposes the Copy action slightly less directly than SwiftUI's `.contextMenu` did (UIKit overlay). Can add an `accessibilityAction` follow-up if wanted.

florian-bellotti:
Follow-up commit 38b166f: replaces the UIKit `UIContextMenuInteraction` bridge from 612cc29 with a pure SwiftUI implementation, per preference for native SwiftUI components.

**Approach**: `.contextMenu(menuItems:preview:)` with a width-matched preview — the bubble's rendered width is measured via `onGeometryChange` and the preview is constrained to it, so the platter wraps identically to the in-list bubble. With an explicit preview, the List-cell dismiss mis-targeting no longer pins the platter at a wrong position for the whole interaction.

**Net effect vs the UIKit bridge**: −90 lines, no snapshot machinery, full VoiceOver support back. Known residual: during dismissal the fading platter drifts toward the cell center for ~100 ms (the underlying OS bug is masked, not fixed) — barely perceptible vs the original 1-2 s freeze, and judged an acceptable trade-off for staying pure SwiftUI.

Verified in the iOS 26.5 simulator (automated long-press/dismiss screenshot bursts + pasteboard check for Copy); `swift test` 177/177.

florian-bellotti:
Follow-up commit 1a5b5ff: eliminates the last visible artifact of the context-menu dismiss bug, still 100% SwiftUI.

**Remaining issue after 38b166f**: the dismissing platter drifted ~100ms toward the List-cell center (the OS re-centers the dismiss preview; nothing in SwiftUI controls that target — verified through iOS 26 APIs).

**Fix**: the preview view now observes `@Environment(\.isPresented)`, which flips to false the moment dismissal starts, and fades itself out. The mis-targeted fly-back animation still runs, but on an invisible platter — the dismissal reads as a clean quick dissolve with the real bubble already in place. Verified with repeated automated dismiss screenshot bursts: no frame shows the bubble at a wrong position anymore. Copy re-verified via pasteboard; swift test 177/177.

Net: pure SwiftUI (`UserBubblePreview`, ~18 lines), no UIKit bridge, full VoiceOver support.

florian-bellotti:
Reverted: the branch is back to 612cc29 (UIKit `UIContextMenuInteraction` implementation) — commits 38b166f and 1a5b5ff (pure-SwiftUI `preview:` variants) were dropped via force-push after on-device testing still showed dismiss artifacts. The comments above describing those two commits are obsolete; 612cc29 is the effective implementation: pixel-stable dismiss verified frame-by-frame in the simulator.

PR #302, opened 2026-07-06
Ask for confirmation before deleting a conversation (#292)

### LINKED ISSUE #292: iOS confirmation before deleting a conversation
## What to build

Swipe-to-delete on a conversation always asks first: a confirmation alert naming the conversation and stating permanence, mirroring the existing workspace-archive confirmation. Failures keep the row visible with a specific error. Applies to workspace and Brain conversation lists, including full swipe.

## Acceptance criteria

- [ ] Swipe-delete presents a confirmation naming the conversation before any deletion; Cancel is side-effect free
- [ ] Full swipe routes through the confirmation rather than bypassing it
- [ ] A failed deletion keeps the session in the list and surfaces an error identifying the conversation (store test); success removes exactly the targeted session
- [ ] The behavior is identical on workspace and Brain conversation lists
- [ ] `cd ios && swift test` passes

## Blocked by

None - can start immediately

---

## Problem Statement

On iOS, swiping to delete a conversation removes it immediately and permanently. A conversation is a persisted agent session — a transcript of prompts, agent reasoning, tool activity, and results that cannot be regenerated. One accidental swipe-and-tap, easy to do while scrolling a list on a phone, destroys it with no confirmation and no recovery. This is inconsistent with the same app's own conventions: archiving a workspace, a far more reversible action, asks for confirmation first. The mismatch teaches users the wrong lesson about which actions are safe. Additionally, when a deletion fails partway (for example during a multi-delete), the user only gets a generic error with no clarity about what was actually removed.

## Solution

Deleting a conversation always asks first. Swiping to delete presents a confirmation alert that names the specific conversation and states that deletion is permanent, mirroring the confirmation pattern the app already uses for archiving workspaces. Confirming deletes the session; cancelling leaves everything untouched. If the deletion fails, the conversation row stays in the list and the user sees the error, so the list never lies about what still exists. This applies uniformly everywhere sessions can be swipe-deleted: workspace conversation lists and Brain conversation lists.

## User Stories

1. As a user, I want a confirmation alert before a conversation is deleted, so that a single accidental swipe cannot destroy an agent transcript.
2. As a user, I want the confirmation to name the conversation I am about to delete, so that I can verify I am deleting the right one.
3. As a user, I want the confirmation to make clear the deletion is permanent, so that I understand the stakes before confirming.
4. As a user, I want Cancel to fully abort the deletion with the row restored to normal, so that backing out is consequence-free.
5. As a user deleting a conversation in a workspace, I want the same confirmation I would get anywhere else in the app, so that destructive actions behave consistently.
6. As a user deleting a Brain conversation, I want the identical confirmation flow, so that the Brain surface is not a loophole around the safeguard.
7. As a user whose deletion fails, I want the conversation to remain in the list, so that the UI reflects reality and I can retry.
8. As a user whose deletion fails, I want to see an error message telling me the deletion did not happen, so that I am not left guessing.
9. As a user who confirms deletion of the conversation I currently have open, I want the app to navigate away cleanly, so that I am not left viewing a deleted session.
10. As a user familiar with the workspace archive confirmation, I want the delete confirmation to look and behave like it, so that the app feels coherent.
11. As a user deleting several conversations in a row, I want a confirmation for each swipe-delete, so that speed never silently overrides safety.

## Implementation Decisions

- Swipe-to-delete no longer deletes directly; it records the targeted conversation and presents a confirmation alert. The alert follows the existing workspace-archive confirmation pattern: it names the conversation (its title, or the app's standard fallback naming for untitled sessions), states that deletion is permanent, and offers a destructive-styled Delete action alongside Cancel.
- Deletion is performed only after explicit confirmation. Cancel dismisses the alert and clears the pending target with no side effects.
- On failure, the conversation remains in the list (no optimistic removal that survives a failed request) and the surface's error presentation shows the failure. The failure message identifies the affected conversation rather than only setting a generic error.
- The confirmation applies to every surface with swipe-delete on sessions: workspace conversation lists and Brain conversation lists. Any shared list component gains the behavior once so the surfaces cannot drift.
- Full swipe (the swipe-all-the-way gesture that normally commits the destructive action) also routes through the confirmation rather than bypassing it.
- No undo, soft-delete, or trash is introduced; confirmation-before-delete is the chosen safeguard.

## Testing Decisions

- Store-testable parts: the deletion flow in the conversation store — that a failed delete leaves the session present in the store's list and surfaces an error describing the failure, and that a successful delete removes exactly the targeted session. Swift Testing cases join the existing conversation store session suite, following its prior art, and assert list contents and error state (external behavior), not call sequences.
- The confirmation dialog itself — presentation on swipe, naming, destructive styling, cancel behavior, full-swipe routing, and parity between workspace and Brain lists — is view-layer and is verified manually on device.

## Out of Scope

- Undo or soft-delete with a grace period — considered and rejected in favor of confirmation, matching the app's existing archive-confirmation convention.
- Backend changes to session deletion, including trash or recovery endpoints.
- Bulk-selection delete UI or changes to how multi-delete is initiated.
- Confirmation for non-destructive session actions.

## Further Notes

The guiding principle is proportionality: the app already confirms workspace archiving, which is reversible, while permanently deleting an irreplaceable transcript was unconfirmed. This PRD closes that inversion using the pattern users already know from the archive flow, so it adds no new interaction vocabulary. The partial-failure improvement is intentionally modest — keep the row, name the conversation in the error — and stops short of a retry queue or reconciliation UI.

florian-bellotti:
## Summary
- Swipe-to-delete on a conversation now shows a confirmation alert naming the conversation and stating the deletion is permanent, mirroring the workspace-archive confirmation pattern in HubView. Cancel is side-effect free.
- Full swipe no longer bypasses confirmation: `.onDelete` is replaced with `.swipeActions(edge: .trailing, allowsFullSwipe: false)`, so every delete routes through the alert. Both the workspace and Brain conversation lists share `ConversationsSection`, so behavior is identical on both.
- The delete flow (non-optimistic removal, focus-fallback, per-session state cleanup) moved from the view into `ConversationStore.deleteSessions(_:from:focusedSessionId:delete:)` with the API call injected as an async throwing closure for testability.
- On failure the session stays in the list and the error names the conversation: `Failed to delete "<title>": <reason>`.
- Added a shared `SessionMetadata.displayTitle` ("Untitled Conversation" fallback) reused by `ConversationRow`, the confirmation message, and the failure error.

## Testing
- `cd ios && swift test` — 121 tests in 18 suites passed (includes 4 new ConversationStore delete-flow tests).
- `xcodebuild -project HiveMobile.xcodeproj -scheme HiveMobile -destination 'generic/platform=iOS Simulator' build` — BUILD SUCCEEDED.

Closes #292

### DIFF
--- a/ios/HiveMobile/Models/Models.swift
+++ b/ios/HiveMobile/Models/Models.swift
@@ -181,6 +181,13 @@ struct SessionMetadata: Codable, Hashable, Identifiable {
 
     var id: String { sessionId }
 
+    var displayTitle: String {
+        guard let title = title?.trimmingCharacters(in: .whitespacesAndNewlines), !title.isEmpty else {
+            return "Untitled Conversation"
+        }
+        return title
+    }
+
     init(
         sessionId: String,
         providerSessionId: String?,
--- a/ios/HiveMobile/Stores/ConversationStore.swift
+++ b/ios/HiveMobile/Stores/ConversationStore.swift
@@ -485,6 +485,54 @@ final class ConversationStore {
         }
     }
 
+    /// Delete sessions via the injected `delete` call (REST in the app, a stub in
+    /// tests) and reconcile store state. Returns the remaining session list and an
+    /// error message naming the conversation when a delete fails.
+    func deleteSessions(
+        _ targets: [SessionMetadata],
+        from sessions: [SessionMetadata],
+        focusedSessionId: String?,
+        delete: (String) async throws -> Void
+    ) async -> (sessions: [SessionMetadata], errorMessage: String?) {
+        var deletedSessionIds = Set<String>()
+        var errorMessage: String?
+        for session in targets {
+            do {
+                try await delete(session.sessionId)
+                deletedSessionIds.insert(session.sessionId)
+                errorMessage = nil
+            } catch is CancellationError {
+                // View disappeared.
+            } catch {
+                errorMessage = "Failed to delete \"\(session.displayTitle)\": \(error.localizedDescription)"
+            }
+        }
+
+        guard !deletedSessionIds.isEmpty else {
+            return (sessions, errorMessage)
+        }
+
+        var remaining = sessions
+        remaining.removeAll { deletedSessionIds.contains($0.sessionId) }
+
+        if let focusedSessionId, deletedSessionIds.contains(focusedSessionId) {
+            let fallbackSessionId = remaining.first?.sessionId
+            if sessionId == nil {
+                setFocusedSessionId(focusedSessionId)
+            }
+            removeSessionState(focusedSessionId, fallbackSessionId: fallbackSessionId)
+            if let fallbackSessionId {
+                _ = await send?(.switchSession(sessionId: fallbackSessionId))
+            }
+        }
+
+        for deletedSessionId in deletedSessionIds where deletedSessionId != focusedSessionId {
+            removeSessionState(deletedSessionId, fallbackSessionId: nil)
+        }
+
+        return (remaining, errorMessage)
+    }
+
     // MARK: - Private
 
     /// Ensure a stream slot exists for the given session, defaulting to streaming state.
--- a/ios/HiveMobile/Stores/ConversationStore.swift
+++ b/ios/HiveMobile/Stores/ConversationStore.swift
@@ -485,38 +485,25 @@ final class ConversationStore {
         }
     }
 
-    /// Delete sessions via the injected `delete` call (REST in the app, a stub in
-    /// tests) and reconcile store state. Returns the remaining session list and an
-    /// error message naming the conversation when a delete fails.
-    func deleteSessions(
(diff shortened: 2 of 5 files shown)


florian-bellotti:
Follow-up from the branch review (commit 0b5c074):

- `ConversationStore.deleteSessions` is now a single-session `deleteSession` returning `(deleted, errorMessage)` — the UI only ever deletes one conversation at a time, and the multi-target loop had a latent path where a success cleared an earlier failure's error message.
- `ConversationsSection` now removes the deleted row from live state instead of overwriting `sessions` with a pre-await snapshot, closing a small race where a refresh or `createSession` landing mid-delete was clobbered.
- Tests adapted to the new signature, plus a new cancellation-path test.

Verified: `cd ios && swift test` (122 tests pass) and `xcodebuild ... build` (BUILD SUCCEEDED).

PR #303, opened 2026-07-06
Fix iOS image attachment reselection, visible errors, and loading state

### LINKED ISSUE #293: iOS image attachment fixes: reselection, visible errors, loading state
## What to build

An honest photo attachment flow: the picker selection clears after each processed pick (fixing the re-pick dead end), a lightweight loading chip acknowledges each pick immediately, and load failures surface a visible inline error near the composer instead of silence.

## Acceptance criteria

- [ ] Removing an attachment chip then re-picking the same photo works (picker selection cleared after processing)
- [ ] A loading chip appears while image data loads and becomes the thumbnail chip on success
- [ ] Load failures show a visible, transient/dismissible inline error near the composer
- [ ] Pending chips are never sent nor persisted to the draft; removing a chip mid-load cancels it cleanly; simultaneous picks load independently
- [ ] `cd ios && swift test` passes (draft attachment persistence unchanged)

## Blocked by

None - can start immediately

---

## Problem Statement

Attaching a photo to a prompt on iOS breaks down in three ways. First, if a user attaches a photo, removes its chip, and then picks the same photo again, nothing happens at all — the picker's selection is only cleared when a message is sent, so re-picking an already-selected item produces no change and the app appears to hang. Second, when loading a picked photo's data fails, the failure is completely silent: the user picks a photo, the picker dismisses, and no chip appears, with no explanation. Third, large photos take a noticeable moment to load and there is no pending indication, so during that gap the flow is indistinguishable from the silent-failure case. Together these make the attachment flow feel unreliable precisely when the user cannot tell whether to wait, retry, or give up.

## Solution

Make the attachment flow honest at every step. The picker selection is cleared as soon as each pick is processed, so removing a chip and re-picking the same photo works every time. While a photo's data is loading, a lightweight placeholder chip appears in the attachment tray immediately, so the user sees instant acknowledgment of the pick; it becomes the real thumbnail chip when loading finishes. If loading fails, the user sees a visible inline error near the composer telling them the image could not be attached, instead of silence. Draft behavior is unchanged: successfully attached images continue to be stored with the per-session draft and restored with it.

## User Stories

1. As a user, I want to re-pick a photo I previously removed from my draft, so that changing my mind about an attachment is not a dead end.
2. As a user, I want the picker selection to reset after each pick is handled, so that every visit to the photo picker behaves like a fresh choice.
3. As a user picking a large photo, I want a pending chip to appear immediately, so that I know the app registered my pick and is working.
4. As a user, I want the pending chip to turn into the image thumbnail when loading completes, so that the transition from loading to attached is obvious.
5. As a user whose photo fails to load, I want a visible error near the composer, so that I know it failed rather than assuming the app is broken.
6. As a user who sees an attachment error, I want it to be dismissible or short-lived, so that it informs me without permanently cluttering the composer.
7. As a user who hits a load failure, I want to be able to immediately try picking the same or another photo, so that recovery is one tap away.
8. As a user attaching several photos at once, I want each photo to load and appear independently, so that one slow or failing image does not block the others.
9. As a user, I want a pending chip that never resolves into a real image to not be sent with my message, so that only successfully loaded images reach the agent.
10. As a user who attaches images and then switches conversations, I want successfully attached images restored with my draft, so that draft behavior stays as reliable as it is today.
11. As a user removing a pending chip before it finishes loading, I want the removal to cancel that attachment cleanly, so that the image does not appear after I removed it.

## Implementation Decisions

- The photo picker's selection array is cleared immediately after each selection change is processed, rather than only on send. This makes every pick produce a fresh change event, fixing the silent no-op when re-picking a previously removed photo.
- Each picked item shows a lightweight loading chip in the attachment tray as soon as processing starts: same footprint as a normal attachment chip, with a placeholder or progress treatment instead of a thumbnail. On success it is replaced in place by the standard thumbnail chip; on failure it is removed.
- Load failures — the transferable data failing to load, the data not decoding into an image, or attachment encoding failing — all surface the same way: a visible inline error message near the composer stating that the image could not be attached. The error is transient or dismissible and does not block further composing or picking.
- Multiple simultaneous picks load independently; one failure produces an error without affecting sibling loads.
- Only successfully loaded attachments participate in sending and in draft persistence. Loading chips are ephemeral UI state: they are never persisted to the draft and never included in a send. Draft storage and restoration of attached images is otherwise unchanged.
- Removing a loading chip cancels or orphans that load so a late-arriving result does not resurrect a removed attachment.

## Testing Decisions

- Store-testable parts: draft attachment persistence remains covered by the existing draft store suite — attachments saved with a draft are restored with it, and cleared on send. If any attachment-state bookkeeping (pending versus loaded, exclusion of pending items from the sendable set) is factored into store- or model-level logic, it gets Swift Testing cases asserting observable outcomes: what is persisted, what is sendable, and what remains after a simulated failure. These follow the prior art of the existing draft store tests and avoid asserting implementation details.
- The picker interaction itself is view-layer and verified manually on device: the remove-then-re-pick-same-photo cycle (the core regression case), the loading chip on a large photo, the inline error on a forced failure, multi-photo picks with mixed outcomes, and removal of a chip mid-load.

## Out of Scope

- Camera capture and file or document attachments.
- Image editing, cropping, or annotation before sending.
- Upload or send progress for the message itself; this PRD covers only the local load-into-draft phase.
- Changes to image resizing, compression, or the attachment format sent to the backend.

## Further Notes

The re-pick fix is the highest-value change because the current behavior is indistinguishable from a hang and has no workaround except picking a different photo first. The loading chip and the visible error are complementary: the chip makes the success path legible and the error makes the failure path legible, and together they eliminate the ambiguous silence that currently follows every pick. Photo library permission-denied states are handled by the system picker itself and are not part of this PRD.

florian-bellotti:
Fixes #293

## Summary

- Clear the PhotosPicker selection immediately after each pick is processed (instead of on send), so removing a chip and re-picking the same photo works every time.
- Show a pending shimmer chip in the attachment tray as soon as a pick starts loading; it is replaced in place by the thumbnail chip on success and removed on failure. Simultaneous picks load independently.
- Show a dismissible inline error banner near the composer ("Image could not be attached.") when the transferable load fails, the data does not decode, or encoding fails.

Pending chips are never sent and never persisted to the draft; removing a pending chip orphans its in-flight load so a late result cannot resurrect it. Draft persistence of successfully attached images is unchanged.

## Acceptance criteria

- [x] Removing an attachment chip then re-picking the same photo works (picker selection cleared after processing)
- [x] A loading chip appears while image data loads and becomes the thumbnail chip on success
- [x] Load failures show a visible, transient/dismissible inline error near the composer
- [x] Pending chips are never sent nor persisted to the draft; removing a chip mid-load cancels it cleanly; simultaneous picks load independently
- [x] `cd ios && swift test` passes (draft attachment persistence unchanged)

## Verification

- `cd ios && swift test`: 117 tests in 18 suites, all passed
- `xcodebuild -project HiveMobile.xcodeproj -scheme HiveMobile -destination 'generic/platform=iOS Simulator' build CODE_SIGNING_ALLOWED=NO`: BUILD SUCCEEDED

### DIFF
--- a/frontend/src/hooks/useSidebarProjectFolders.ts
+++ b/frontend/src/hooks/useSidebarProjectFolders.ts
@@ -112,12 +112,21 @@ export function useSidebarProjectFolders(
       }
     }
 
+    // Server data (fetches and PUT echoes) must not clobber local edits made
+    // while the request was in flight; those still need their own flush.
+    const wasHydrated = hydratedRef.current;
+    const lastFlushed = lastFlushedRef.current;
     hydratedRef.current = true;
     setHasInitialHydration(true);
     lastFlushedRef.current = serverState;
     bootstrappedRef.current = true;
     setState((prev) => {
-      return areSidebarPreferencesStatesEqual(prev, serverState) ? prev : serverState;
+      if (areSidebarPreferencesStatesEqual(prev, serverState)) return prev;
+      if (wasHydrated && lastFlushed && !areSidebarPreferencesStatesEqual(prev, lastFlushed)) {
+        // Fresh identity so the flush effect re-runs and pushes the kept edits.
+        return { ...prev };
+      }
+      return serverState;
     });
   }, [preferencesData, projectsReady]);
 
--- a/frontend/tests/components/Sidebar.test.tsx
+++ b/frontend/tests/components/Sidebar.test.tsx
@@ -1115,6 +1115,40 @@ describe("Sidebar", () => {
     });
   });
 
+  it("keeps a folder deletion made while the create flush is in flight", async () => {
+    const user = userEvent.setup();
+    renderSidebar("/projects", projects);
+
+    let resolveCreatePut: (() => void) | null = null;
+    vi.mocked(api.put).mockImplementation((url: string, body?: unknown) => {
+      if (url !== "/api/ui-preferences") throw new Error(`Unexpected PUT: ${url}`);
+      if (resolveCreatePut) return Promise.resolve(body);
+      return new Promise((resolve) => {
+        resolveCreatePut = () => resolve(body);
+      });
+    });
+
+    await screen.findByText(withTextContent("acme/alpha"));
+    await user.click(screen.getByRole("button", { name: "New folder" }));
+    await user.type(screen.getByLabelText("Folder name"), "Temporary");
+    await user.click(screen.getByRole("button", { name: "Create folder" }));
+
+    await screen.findByRole("button", { name: "Temporary" });
+    await waitFor(() => expect(api.put).toHaveBeenCalledTimes(1));
+
+    fireEvent.click(screen.getByRole("button", { name: "Delete folder Temporary" }));
+    await user.click(await screen.findByRole("button", { name: "Delete" }));
+
+    await act(async () => {
+      resolveCreatePut?.();
+    });
+
+    await waitFor(() => {
+      expect(screen.queryByRole("button", { name: "Temporary" })).not.toBeInTheDocument();
+    });
+    await waitFor(() => expect(api.put).toHaveBeenCalledTimes(2));
+  });
+
   it("hides the delete button for non-empty folders", async () => {
     const queryClient = new QueryClient({
       defaultOptions: { queries: { retry: false }, mutations: { retry: false } },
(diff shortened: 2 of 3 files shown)


florian-bellotti:
Follow-up commit faa99a3: sending while an attachment was still loading silently dropped it (`handleSend` cleared the pending chip before `loadTransferable` completed). `canSend` now blocks while any attachment is pending, so the send button greys out until the image is ready. Also enlarged the error banner's dismiss tap target to 44pt and added a VoiceOver label. Verified with `swift test` (117 pass) and `xcodebuild build` (BUILD SUCCEEDED).

Issue #285, opened 2026-07-02
iOS Hub server-unreachable state with retry


florian-bellotti:
## What to build

An honest Hub when the server is unreachable: a dedicated "can't reach your Hive server" state (Retry + Open Settings) replaces the misleading "No Projects" empty state when the fetch failed, failure classification (unreachable vs invalid token) lives in the store, and a failed pull-to-refresh with cached data surfaces transient feedback instead of silence.

## Acceptance criteria

- [ ] Empty project list + failed fetch renders the unreachable state with Retry and Open Settings; the No Projects onboarding state only renders after a successful fetch
- [ ] The store classifies failures (unreachable / authentication / other) as observable state, with failure-appropriate copy (auth failures point at Settings)
- [ ] Retry drives the standard fetch path with visible in-flight state; success clears all failure state
- [ ] Pull-to-refresh failure with cached data shows a lightweight transient notice while cached content stays visible
- [ ] `cd ios && swift test` passes with the state-selection matrix covered

## Blocked by

None - can start immediately

---

## Problem Statement

When the iOS app cannot reach the Hive server, the Hub screen actively misleads the user. If the initial project fetch fails and no cached data exists, the Hub falls back to the "No Projects — tap + to add your first project" empty state — a call to action that is factually wrong and potentially harmful: the user's projects exist, the server is just unreachable, and tapping + leads them down a pointless creation flow. The only hint of trouble is a transient error capsule that offers no dismiss and no retry, then disappears. The failure modes with cached data are worse in a different way: background refreshes and even user-initiated pull-to-refresh can fail completely silently, so the user looks at stale project, workspace, and session data while believing it is fresh. There is no honest "we can't reach your server" moment anywhere in the Hub, and no path to recovery other than force-quitting or wandering into Settings on a hunch.

## Solution

Make the Hub tell the truth about reachability. When the project list is empty and the last fetch failed, the Hub shows a dedicated "can't reach your Hive server" state instead of the misleading No Projects empty state — with a prominent Retry button that re-attempts the fetch and an Open Settings shortcut for cases where the server address or credentials are the real problem. The store classifies failures so the message can be accurate: a server that is unreachable reads differently from an authentication failure, which points the user at Settings rather than at their network. When the user explicitly pulls to refresh and the refresh fails but cached data exists, the app surfaces lightweight, transient feedback acknowledging the failure instead of silently keeping stale data on screen — the cached content remains visible, but the user knows it did not just update.

## User Stories

[...]

PR #307, opened 2026-07-06
Debounce the connection banner and drop the connecting toast (iOS)

florian-bellotti:
Fixes the connection-banner UX regressions reported on `main` (from #304).

## Problem
- The red "Disconnected. Tap to reconnect" pill was shown **instantly with no grace**, so while the WebSocket hub connection flaps (REST works, WS drops/reconnects on backoff) the button **strobed on and off in a loop**.
- The orange "Connecting…" toast appeared on **every app launch and wake-from-sleep** — noise, not signal.

## Change (iOS only)
- `ConnectionBanner`: removed the "Connecting…" state entirely. The disconnected pill now only appears after the connection has been **continuously disconnected for ~3s** (`.task(id: connectionState)` restarts on every state change), so cold start, wake, and transient flaps surface nothing; a genuine sustained outage shows a steady pill. Placement (`safeAreaInset` below the nav bar) unchanged.
- `HubStatusMonitor`: applied plan 005 — extracted `ensureConnection()` and made `reconnectNow()` self-healing (rebuilds a connection when none exists instead of a silent no-op), so the reconnect button can never be dead.

## Tests
- Added `reconnectNowIsSafeNoOpWhenNothingSubscribed` to `HubStatusMonitorTests`.
- `cd ios && swift test` → 186 pass. `xcodebuild build -scheme HiveMobile -destination 'platform=iOS Simulator,name=iPhone 17 Pro' CODE_SIGNING_ALLOWED=NO` → BUILD SUCCEEDED.

Executes the previously-written `plans/005`; the placement/transition (`plans/006`) was already applied in #304.

## Manual verification
Screenshot to follow once the sustained-disconnect state can be captured.

### DIFF
--- a/ios/HiveMobile/Stores/HubStatusMonitor.swift
+++ b/ios/HiveMobile/Stores/HubStatusMonitor.swift
@@ -175,6 +175,13 @@ final class HubStatusMonitor {
 
     // MARK: - Sync
 
+    private func ensureConnection() {
+        guard hubConnection == nil, !currentSyncPayload.workspaceIds.isEmpty else { return }
+        let conn = makeHubConnection(self)
+        hubConnection = conn
+        conn.connect()
+    }
+
     /// Sync monitored workspaces — manages hub connection and subscriptions.
     func sync(workspaceIds: [String]) {
         let desired = Set(workspaceIds)
@@ -198,9 +205,7 @@ final class HubStatusMonitor {
             hubConnection = nil
             connectionState = .disconnected
         } else if hubConnection == nil {
-            let conn = makeHubConnection(self)
-            hubConnection = conn
-            conn.connect()
+            ensureConnection()
         } else if let (_, payload) = change {
             hubConnection?.sendSync(payload)
         }
@@ -255,7 +260,11 @@ final class HubStatusMonitor {
     }
 
     func reconnectNow() {
-        hubConnection?.forceReconnect()
+        if let hubConnection {
+            hubConnection.forceReconnect()
+        } else {
+            ensureConnection()
+        }
     }
 
     /// Workspaces that were streaming when the app entered background.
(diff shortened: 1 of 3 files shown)


florian-bellotti:
## Manual test — connection banner ✅

Ran the branch on the iPhone 17 Pro simulator.

**1. Connecting / connected — no more "Connecting…" toast.** On launch the Hub loads with **no banner at all** (the orange connecting toast is gone):

No connecting toast

**2. Disconnected — debounced "Tap to reconnect" pill.** After a sustained (>3 s) loss of connection, the red pill appears and taps to reconnect:

Disconnected pill

Verified: no orange connecting toast on launch, no strobe/flicker, and the disconnected pill only surfaces after the 3 s debounce (the "No Projects" / network-error copy in the second shot is `main`'s pre-#285 empty state, unrelated to this PR).

florian-bellotti:
Follow-up from device testing:

- **First-launch false "Disconnected. Tap to reconnect"** — the banner now only appears once the hub has connected **at least once** (persisted `hasEverConnected`). A never-configured first launch no longer shows a disconnect state.
- **"Message not sent: disconnected from server" while Settings shows connected** — the send path (WS) and the health state were decoupled. On a workspace send failure we now probe the socket, so the connection state + banner reconcile with reality and reconnect if it silently dropped.

florian-bellotti:
Round 2 (device testing):
- **Send failures** — a workspace send that hit a down/absent socket was dropped as "not sent". Now the send path (re)connects and **retries once** within a short window, so a send during a brief disconnect actually goes through.
- **Honest banner** — the disconnect banner was gated on `hasEverConnected`, which *hid* the exact confusing case (Settings shows connected via REST, but the hub socket is down). Now gated on whether a server is configured: never-configured launches stay clean (onboarding handles them); a configured-but-disconnected session honestly shows the banner.

florian-bellotti:

PR #309, opened 2026-07-06
Honor Reduce Motion in iOS activity animations (#296)

### LINKED ISSUE #296: iOS Reduce Motion support
## What to build

Respect the system Reduce Motion setting: the dot-grid activity indicator, pulsing streaming dot, and repeat-forever shimmer/pulse effects render static equivalents (skipping the animation work entirely) while keeping activity, streaming, and loading states visually distinct. Default experience unchanged.

## Acceptance criteria

- [ ] Reduce Motion is read from the standard accessibility environment; no app-level duplicate setting
- [ ] With the setting on: static dot grid that still reads as active, steady streaming dot, static shimmer placeholders — with the per-frame animation work skipped, not hidden
- [ ] The pattern is centralized in the shared indicator/shimmer components so all call sites inherit it
- [ ] Toggling the setting mid-session switches presentation immediately; with the setting off, existing animations are unchanged
- [ ] `cd ios && swift test` passes (no store behavior changes)

## Blocked by

- #283 (both touch the animated activity indicators; land the 30 Hz cap first to avoid conflicts)

---

## Problem Statement

Users who enable Reduce Motion in iOS accessibility settings do so because animation causes them discomfort — vestibular disorders, motion sensitivity, or simple distraction. Hive never reads that setting. Its most prominent visual element, the animated dot-grid agent-activity indicator, runs continuously at display refresh rate and can appear many times at once — in Hub rows, conversation rows, the chat header, and dashboards. A streaming dot pulses continuously, and shimmer and pulse effects repeat forever. For a motion-sensitive user, the busiest screens of the app are exactly the ones they cannot comfortably look at, and the system-wide setting they rely on is silently ignored.

## Solution

The app respects Reduce Motion. When the setting is enabled, every continuous animation is replaced with a static equivalent that conveys the same information: the dot-grid activity indicator renders as static dots in a fixed state that still reads as "agent working", the pulsing streaming dot becomes a steady dot, and repeat-forever shimmer and pulse effects render as static placeholders. One-shot transitions (view appearance, navigation) may remain but prefer opacity changes over movement. No information is lost — activity, streaming, and loading states stay visually distinct — only the motion is removed.

## User Stories

1. As a motion-sensitive user with Reduce Motion enabled, I want the dot-grid agent-activity indicator to render as static dots, so that I can look at the Hub without discomfort.
2. As a user with Reduce Motion enabled viewing the Hub, I want a screen full of active workspaces to be completely still, so that many simultaneous indicators do not multiply into an overwhelming surface.
3. As a user with Reduce Motion enabled in a chat with a working agent, I want the streaming indicator to be steady instead of pulsing, so that I can read the incoming response without a flashing element beside it.
4. As a user with Reduce Motion enabled, I want shimmer loading effects replaced with static placeholders, so that loading screens do not strobe.
5. As a user with Reduce Motion enabled, I want to still be able to tell that an agent is active, so that removing motion does not remove information.
6. As a user with Reduce Motion enabled, I want streaming, idle, and completed states to remain visually distinguishable, so that the static presentation is as informative as the animated one.
7. As a user who toggles Reduce Motion while the app is running, I want the app to switch presentation immediately, so that I do not need to relaunch.
8. As a user with Reduce Motion enabled, I want screen transitions to favor simple fades over sliding or scaling movement, so that navigation feels calm.
9. As a user without Reduce Motion enabled, I want all existing animations to remain exactly as they are, so that the default experience is unchanged.
10. As a user with Reduce Motion enabled, I want the conversation rows and dashboard views to be as still as the Hub, so that the setting applies consistently everywhere the activity indicator appears.

## Implementation Decisions

- Read the system reduce-motion state via the standard SwiftUI accessibility environment value and branch presentation on it; no app-level setting duplicates the system one.
- Dot-grid agent-activity indicator: when Reduce Motion is on, render the dots at a fixed, deterministic state that visually reads as active; do not run the per-frame animation loop at all (skipping the work entirely, not just hiding it).
- Streaming dot: static filled dot instead of a pulse.
- Shimmer and other repeat-forever effects: render the static base placeholder with no animation pass.
- One-shot animations and transitions may remain, but where a transition involves movement, prefer an opacity-based variant under Reduce Motion.
- Centralize the pattern where practical (for example, the shared indicator and shimmer components check the setting internally) so every existing and future call site inherits the behavior without per-screen work.
- This PRD covers honoring the setting only; capping the frame rate of the animated indicators for users without Reduce Motion belongs to the streaming-render performance PRD.

## Testing Decisions

- This is view-layer work: the presence or absence of motion is verified manually by toggling Reduce Motion in system settings and sweeping the Hub, Brain conversation list, chat (streaming and idle), and dashboard screens, including toggling the setting mid-session.
- Any small pure logic that emerges — for example, a deterministic static-state computation for the dot grid — can be asserted with Swift Testing in the existing iOS test suite alongside the store tests, asserting external behavior rather than implementation details.
- `cd ios && swift test` is the automated check for the stores; no store behavior should change under this PRD. The app is iPhone-only.

## Out of Scope

- Frame-rate capping of the animated indicators for the default (motion-enabled) experience — handled by the streaming-render performance PRD.
- Any new animations.
- Reduce Transparency support.
- Changes to the web frontend.

## Further Notes

- Skipping the animation work entirely under Reduce Motion (rather than animating invisibly) also yields a battery and CPU benefit for these users, which complements the separate performance PRD without overlapping it.
- The static "active" state for the dot grid should be designed once and reused everywhere the indicator appears, so the reduced-motion look is consistent across Hub rows, conversation rows, chat, and dashboards.

florian-bellotti:
Closes #296.

## What
The app never read the system **Reduce Motion** setting, so continuous animations ran regardless. Each of the three shared animated components now reads `@Environment(\.accessibilityReduceMotion)` internally and renders a static equivalent when Reduce Motion is on. Every call site inherits the behavior — no per-screen work, no app-level setting.

| Component | Motion on (unchanged) | Reduce Motion on |
|---|---|---|
| `ShimmerModifier` | opacity oscillates via `repeatForever` | fixed 0.5 opacity, no animation pass |
| `PulseModifier` | scale pulses to 1.02 | stays at scale 1.0, no animation |
| `ChatActivityRowLabel` streaming dot | `TimelineView` opacity pulse | steady filled dot |
| `AgentActivityIndicator` | `TimelineView` wave over the dot grid | static `canvas(t: 0)` frame (reads as "working") |

Same information conveyed, no motion. The default (motion-on) path is byte-for-byte unchanged.

## Scope
- iOS only. Files: `Theme/Animations.swift`, `Views/Chat/ChatActivityChrome.swift`, `Views/Components/AgentActivityIndicator.swift`.
- Out of scope (unchanged): the `TimelineView` frame-rate cap from #283, Reduce Transparency, any new animations.

## Testing
- `cd ios && swift test` → all pass (185 tests).

## Screenshots
_UI/behavior change. A still screenshot of a paused vs. static animation looks identical, so before/after captures of the static Reduce-Motion states (Settings → Accessibility → Motion → Reduce Motion ON) will be attached from the simulator for the record._

### DIFF
--- a/ios/HiveMobile/Theme/Animations.swift
+++ b/ios/HiveMobile/Theme/Animations.swift
@@ -9,10 +9,13 @@ struct ShimmerModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
             .opacity(reduceMotion ? 0.5 : 0.5 + 0.5 * Foundation.sin(phase))
-            .onAppear {
-                guard !reduceMotion else { return }
-                withAnimation(.easeInOut(duration: 1.5).repeatForever(autoreverses: true)) {
-                    phase = .pi
+            .onChange(of: reduceMotion, initial: true) {
+                if reduceMotion {
+                    phase = 0
+                } else {
+                    withAnimation(.easeInOut(duration: 1.5).repeatForever(autoreverses: true)) {
+                        phase = .pi
+                    }
                 }
             }
     }
--- a/ios/HiveMobile/Theme/Animations.swift
+++ b/ios/HiveMobile/Theme/Animations.swift
@@ -3,12 +3,14 @@ import SwiftUI
 // MARK: - Shimmer (loading skeletons)
 
 struct ShimmerModifier: ViewModifier {
+    @Environment(\.accessibilityReduceMotion) private var reduceMotion
     @State private var phase: CGFloat = 0
 
     func body(content: Content) -> some View {
         content
-            .opacity(0.5 + 0.5 * Foundation.sin(phase))
+            .opacity(reduceMotion ? 0.5 : 0.5 + 0.5 * Foundation.sin(phase))
             .onAppear {
+                guard !reduceMotion else { return }
                 withAnimation(.easeInOut(duration: 1.5).repeatForever(autoreverses: true)) {
                     phase = .pi
                 }
@@ -20,13 +22,16 @@ struct ShimmerModifier: ViewModifier {
 
 struct PulseModifier: ViewModifier {
     let isActive: Bool
+    @Environment(\.accessibilityReduceMotion) private var reduceMotion
     @State private var isPulsing = false
 
     func body(content: Content) -> some View {
         content
-            .scaleEffect(isActive && isPulsing ? 1.02 : 1.0)
+            .scaleEffect(!reduceMotion && isActive && isPulsing ? 1.02 : 1.0)
             .animation(
-                isActive ? .easeInOut(duration: 2).repeatForever(autoreverses: true) : .default,
+                reduceMotion
+                    ? nil
+                    : (isActive ? .easeInOut(duration: 2).repeatForever(autoreverses: true) : .default),
                 value: isPulsing
             )
             .onChange(of: isActive) { _, active in
--- a/ios/HiveMobile/Views/Chat/ChatActivityChrome.swift
+++ b/ios/HiveMobile/Views/Chat/ChatActivityChrome.swift
@@ -9,6 +9,7 @@ struct ChatActivityStats {
 }
 
 struct ChatActivityRowLabel: View {
+    @Environment(\.accessibilityReduceMotion) private var reduceMotion
     var icon: String? = nil
     let label: String
     var detail: String?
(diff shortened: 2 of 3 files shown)


florian-bellotti:

PR #310, opened 2026-07-06
Navigate to the workspace on push notification tap (#294)

### LINKED ISSUE #294: iOS push notification tap navigates to the workspace
## What to build

Tapping a "turn completed" push navigates to the workspace it concerns via the existing pending-navigation path (into the session when the payload identifies it), with a refresh-and-retry resolution fallback and a silent Hub fallback for stale payloads. Works from warm resume and cold launch.

## Acceptance criteria

- [ ] Tapping the notification switches to the Hub tab and pushes the resolved workspace via the existing pending-navigation mechanism
- [ ] When the payload identifies an existing session, navigation continues into that session's chat
- [ ] An unresolved workspace triggers one project-list refresh and retry, then falls back silently to the Hub (store test for resolution logic)
- [ ] Cold launch defers navigation until stores are ready; the completed indicator on the Hub row is still set
- [ ] `cd ios && swift test` passes

## Blocked by

None - can start immediately

---

## Problem Statement

When an agent finishes a turn, Hive sends a push notification saying the work is done. But tapping that notification only opens the app wherever it was last left — the Hub, the Brain tab, Settings, or deep inside a different workspace. The only visible effect of the tap is a small completed indicator on the workspace's Hub row. The user then has to remember which workspace the notification was about, switch to the Hub tab, scroll to find the workspace, and open it manually. For the most common mobile flow — "the agent is done, go look at the result" — the notification does not actually take you anywhere, which defeats the purpose of tapping it.

## Solution

Tapping a "turn completed" notification takes the user directly to the relevant workspace. The app resolves the workspace identified in the notification payload, switches to the Hub tab, and pushes the workspace's conversation list onto the navigation stack — reusing the pending-navigation mechanism the app already uses elsewhere for programmatic navigation. If the payload also identifies the session, navigation continues one level deeper into that session's chat. If the workspace cannot be resolved (for example it was archived or deleted since the notification was sent), the app falls back gracefully to the Hub with no error, where the user still sees current state.

## User Stories

1. As a user who receives a "turn completed" push notification, I want tapping it to open the workspace the agent worked in, so that I can review the result immediately.
2. As a user whose notification payload identifies the specific session, I want to land directly in that session's chat, so that I do not have to pick the right conversation from a list.
3. As a user who was on the Brain tab when I tapped the notification, I want the app to switch to the Hub tab automatically, so that I am not left in an unrelated context.
4. As a user who was already deep in another workspace, I want the notification tap to navigate to the notified workspace anyway, so that the tap always means "take me there".
5. As a user launching the app cold from a notification tap, I want navigation to happen once the project list has loaded, so that a cold start behaves the same as a warm one.
6. As a user whose workspace list is stale, I want the app to refresh the project list before giving up on resolving the workspace, so that recently created workspaces are still reachable from notifications.
7. As a user who taps a notification for a workspace that has since been archived or deleted, I want to land on the Hub with no error dialog, so that a stale notification never feels like a crash or a dead end.
8. As a user who taps several queued notifications in a row, I want the app to navigate to the most recent tap's target, so that I end up where I last asked to go rather than in a confusing navigation pile-up.
9. As a user who dismisses a notification without tapping it, I want the existing behavior — the completed indicator on the Hub row — to remain, so that I still get the passive signal without forced navigation.
10. As a user, I want the completed indicator on the Hub row to still be set when I navigate via the notification, so that Hub state stays consistent with what actually happened.

## Implementation Decisions

- On notification tap, extract the workspace identifier (and session identifier when present) from the notification payload.
- Resolve the workspace against the loaded project list. If the workspace is not found, force a refresh of the project list first, then retry resolution once.
- On successful resolution, trigger the existing pending-navigation path: it already switches the selected tab to Hub and pushes the workspace onto the Hub navigation stack. Do not build a parallel navigation mechanism.
- If the payload identifies the session and the session exists in that workspace, continue navigation into the session's conversation after the workspace push.
- If resolution fails after refresh (archived, deleted, or unknown workspace), fall back to the Hub tab root silently — no alert, no toast.
- Preserve the existing side effect of marking the workspace's completed indicator; navigation is additive, not a replacement.
- Handle both app states: warm foreground/background resume and cold launch from the notification, deferring navigation until stores are ready in the cold-launch case.

## Testing Decisions

- Testable logic lives in the stores: navigation-target resolution (payload in, resolved workspace/session or fallback out), including the "refresh then retry" path and the "unknown workspace falls back silently" path, belongs on the project store and is tested with Swift Testing under the existing iOS test suite (prior art: HubOrganizationTests, ConversationStoreSessionTests). Tests assert external behavior — what target gets resolved and what pending navigation is requested — not implementation details.
- The actual tab switch, navigation-stack push, and cold-launch timing are view-layer and verified manually on device: tap from foreground, from background, from a cold launch, and with a stale payload.
- `cd ios && swift test` is the automated check. The app is iPhone-only.

## Out of Scope

- Notification actions (approving a plan or replying to the agent from the notification itself).
- Deep links from outside the app (custom URL scheme, universal links).
- Notification settings or per-workspace notification preferences.
- Changes to what notifications are sent or their content.

## Further Notes

- The pending-navigation mechanism is already exercised by other flows, so this PRD is mostly wiring the notification tap into a proven path; the main new logic is payload resolution with the refresh-and-retry fallback.
- Session-level continuation should degrade gracefully: if the session identifier is missing or stale but the workspace resolves, landing on the workspace's conversation list is the correct outcome, not a failure.

florian-bellotti:
Closes #294.

## What
Tapping a "turn completed" push previously only marked the Hub completion dot and left the user wherever they were. It now **navigates to the workspace** via the existing `pendingNavigation` mechanism (the same path used after workspace creation) — no parallel navigation system introduced.

## Behavior
- **Tap → navigate**: `AppDelegate` forwards `workspaceId` (and `sessionId` if the payload carries it) as a navigation request, while still marking the completed indicator.
- **Resolution**: `ProjectStore.resolvePushTarget` finds the workspace in loaded projects; if absent it refreshes the project list once and retries; if still unresolved (archived/deleted/unknown) it returns nil.
- **Silent fallback**: an unresolved target lands the user on the Hub root — no alert.
- **Cold launch**: handled in `onAppear` (the resolve awaits the project refresh before navigating); warm taps handled via `onChange`.
- **Session continuation**: a targeted session is pushed once its `ConversationsSection` loads; a stale/absent session degrades gracefully to the conversation list.

> Note: the backend `agent_turn_complete` payload currently carries only `workspaceId`, so the session-continuation path is plumbed but dormant until a `sessionId` is added server-side. It degrades correctly today.

## Testability
`ProjectStore`'s project/preferences fetches are now injectable (optional init params; production defaults unchanged), so the resolution logic is unit-tested without the network.

## Scope
iOS only. `HiveApp.swift`, `Services/AppDelegate.swift`, `Services/CompletedWorkspacesStore.swift`, `Stores/ProjectStore.swift`, `Views/Chat/ConversationsSection.swift`, `Package.swift` (adds `ProjectStore`/`APIClient` to the test target), plus a new test file.

## Testing
- `cd ios && swift test` → **189 tests pass** (+4 new: loaded, refresh-retry, unresolved-fallback, cold-launch).

## Screenshots
_This is a navigation-flow change (tap push → land on the workspace) rather than a static visual change; a screen recording of the tap-to-workspace flow will be captured from the simulator for the record._

### DIFF
--- a/ios/HiveMobile/HiveApp.swift
+++ b/ios/HiveMobile/HiveApp.swift
@@ -76,10 +76,15 @@ struct HiveApp: App {
             }
             .onAppear {
                 mergePushCompletions()
+                handlePushNavigation()
             }
             .onChange(of: CompletedWorkspacesStore.shared.pending) { _, pending in
                 guard !pending.isEmpty else { return }
                 mergePushCompletions()
+            }
+            .onChange(of: CompletedWorkspacesStore.shared.navigationRequest) { _, request in
+                guard request != nil else { return }
+                handlePushNavigation()
             }
             .onChange(of: scenePhase) { _, newPhase in
                 switch newPhase {
@@ -108,6 +113,27 @@ struct HiveApp: App {
             projectStore.statusMonitor.markCompletedFromPush(wsId)
         }
         store.clearAll()
+    }
+
+    /// Resolve the most recent push tap and navigate to its workspace (and session),
+    /// falling back silently to the Hub tab when the workspace can't be resolved.
+    private func handlePushNavigation() {
+        guard let request = CompletedWorkspacesStore.shared.navigationRequest else { return }
+        CompletedWorkspacesStore.shared.clearNavigationRequest()
+        Task { @MainActor in
+            if let target = await projectStore.resolvePushTarget(
+                workspaceId: request.workspaceId,
+                sessionId: request.sessionId
+            ) {
+                projectStore.pendingSessionNavigation = target.sessionId.map {
+                    PendingSessionNavigation(workspaceId: target.workspace.id, sessionId: $0)
+                }
+                projectStore.pendingNavigation = target.workspace
+            } else {
+                projectStore.pendingSessionNavigation = nil
+                selectedTab = .hub
+            }
+        }
     }
 }
 
--- a/ios/HiveMobile/Services/AppDelegate.swift
+++ b/ios/HiveMobile/Services/AppDelegate.swift
@@ -39,16 +39,17 @@ final class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCent
         return []
     }
 
-    /// Handle notification tap — mark workspace as completed so the completion dot appears.
+    /// Handle notification tap — mark the workspace completed and request navigation to it.
     func userNotificationCenter(
         _ center: UNUserNotificationCenter,
         didReceive response: UNNotificationResponse
     ) async {
         let userInfo = response.notification.request.content.userInfo
-        if let workspaceId = userInfo["workspaceId"] as? String {
-            await MainActor.run {
-                CompletedWorkspacesStore.shared.insert(workspaceId)
-            }
+        guard let workspaceId = userInfo["workspaceId"] as? String else { return }
+        let sessionId = userInfo["sessionId"] as? String
+        await MainActor.run {
+            CompletedWorkspacesStore.shared.insert(workspaceId)
+            CompletedWorkspacesStore.shared.requestNavigation(workspaceId: workspaceId, sessionId: sessionId)
         }
     }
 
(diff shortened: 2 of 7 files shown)


florian-bellotti:

PR #313, opened 2026-07-07
Fix composer model picker, duplicate images, and remove-image control (iOS)

florian-bellotti:
From device testing of the integrated build.

## Fixes
- **Model button did nothing** — `ModelCatalog.loadIfNeeded()` guarded on `isLoaded` and never retried. If the catalog failed to load at launch (server unreachable on first run), it stayed empty forever, so the model `Menu` had no items and tapping it did nothing. The catalog now reloads when a conversation appears (with an in-flight guard against concurrent loads).
- **Same image could be uploaded twice** — image attachments are now de-duplicated by data URL.
- **Remove-image button looked wrong** — replaced the doubled circle (a white glyph on a near-white disc in light mode) with a crisp palette `xmark.circle.fill` (white X on a translucent dark disc), plus an accessibility label.

## Notes
- The reported *input bar too high / breaks behind the keyboard on first open* is a separate root cause (the tab-bar hide transition) and is handled in the navigation PR.
- iOS-only. `swift test` green (186), `xcodebuild` green, 0 warnings.

### DIFF
--- a/ios/HiveMobile/Stores/ModelCatalog.swift
+++ b/ios/HiveMobile/Stores/ModelCatalog.swift
@@ -15,15 +15,18 @@ final class ModelCatalog {
     private(set) var models: [ModelCatalogEntry] = []
     private(set) var defaultModelId: String = ""
     private(set) var isLoaded = false
+    private var isLoading = false
 
     private let api = APIClient()
 
     func loadIfNeeded() async {
-        guard !isLoaded else { return }
+        guard !isLoaded, !isLoading else { return }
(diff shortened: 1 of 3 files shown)


florian-bellotti:
Round 2 (device testing): the model menu appended a trailing checkmark only to the selected row, shifting the rows. Replaced with a native inline **Picker** so selection uses the system's aligned checkmark; when a session is locked to a provider, only that provider's models are listed.

florian-bellotti:

PR #314, opened 2026-07-07
Smooth the bottom tab bar transition in and out of conversations (iOS)

florian-bellotti:
From device testing.

## Problem
- The bottom tab bar (Brain/Hub/Settings) reappeared with a **multi-second delay** when navigating back from a conversation to the workspace list.
- On **first open** of a conversation the composer input bar sat **too high** and could end up behind the keyboard — the bottom safe area still reserved the tab bar's height until its hide animation settled. Leaving and returning fixed it (tab bar already hidden the second time).

## Fix
Bind tab bar visibility to each `NavigationStack`'s **path depth** on the stable stack view (`.toolbar(path.isEmpty ? .automatic : .hidden, for: .tabBar)`) instead of hiding it per pushed destination. Visibility now flips together with the push/pop transition: hidden *before* the conversation lays out (so the composer's safe area is correct on first frame), and restored immediately when the stack pops to its root.

## Verification
- `xcodebuild` green, 0 warnings; store tests unaffected.
- ⚠️ This is a navigation/animation change I could **not** visually verify on-simulator (no UI-automation tap available in this environment). Please confirm the transition feels right on your device — this is the 'keep hidden but smooth' approach you picked.

### DIFF
--- a/ios/HiveMobile/HiveApp.swift
+++ b/ios/HiveMobile/HiveApp.swift
@@ -38,6 +38,7 @@ struct HiveApp: App {
                             navigationPath: $brainPath
                         )
                     }
+                    .toolbar(brainPath.isEmpty ? .automatic : .hidden, for: .tabBar)
                 }
                 Tab("Hub", systemImage: "square.grid.2x2.fill", value: .hub) {
                     NavigationStack(path: $hubPath) {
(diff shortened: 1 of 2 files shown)


florian-bellotti:

PR #312, opened 2026-07-06
Honest Hub server-unreachable state with retry (#285)

florian-bellotti:
Closes #285.

## What
When the project fetch fails and there's no cached data, the Hub showed a misleading **"No Projects"** onboarding state. It now shows an honest **"can't reach your Hive server"** state with **Retry** and **Open Settings**, and a failed pull-to-refresh over cached data surfaces a lightweight transient notice instead of silence.

## Behavior
- **Failure classification in the store** (`ProjectStore`): `unreachable` (transport/URLError), `authentication` (401/403), `other` (5xx, decoding) — observable state, not just a string. Any successful fetch clears it.
- **Honest empty state**: `fetchFailure != nil` → unreachable state; otherwise loading, or the "No Projects" onboarding **only after a fetch has actually succeeded** (`hasLoadedSuccessfully`). Onboarding can no longer masquerade as a failed fetch.
- **Failure-appropriate copy**: auth failures point the user at Settings.
- **Retry** drives the standard forced fetch with a visible in-flight state (spinner + disabled button); success clears all failure state.
- **Pull-to-refresh over cache**: an explicit (`userInitiated`) refresh that fails while cached projects are visible shows a transient top notice ("Showing cached data") that auto-dismisses; background refresh failures stay silent.

## Scope
iOS only. `Stores/ProjectStore.swift`, `Views/Hub/HubView.swift`, `HiveApp.swift` (wires Open Settings via the existing tab nav), `Package.swift` (adds `ProjectStore`/`APIClient` to the test target), new test file.

## Testing
- `cd ios && swift test` → **194 tests pass** (+9: unreachable/auth/other classification, onboarding-not-failure, cached-failure notice + acknowledge, background-silent, success-clears, in-flight-loading, retry-preserves-failure).
- `xcodebuild build … -scheme HiveMobile -destination 'platform=iOS Simulator,name=iPhone 17 Pro'` → **BUILD SUCCEEDED**.

## ⚠️ Merge ordering
This branch and **#294 (PR #310)** both make `ProjectStore` testable — they independently add optional fetch closures to `ProjectStore.init` and add `ProjectStore`/`APIClient` to `Package.swift`'s test target. **Merge one, then rebase the other** (the conflict is mechanical: reconcile the two init signatures into one and keep the single Package.swift additions).

## Screenshots
_UI change — captures of the unreachable state (Retry + Open Settings) and the transient cached-data notice will be attached from the simulator._

PR #311, opened 2026-07-06
Add sparse haptic feedback for key actions (#297)

florian-bellotti:
Closes #297.

## What
Restrained, purposeful haptics for the moments that matter — sparse by design (never a buzz per event), firing only for events on the screen the user is viewing, and automatically respecting the system haptics setting (native `UIFeedbackGenerator` / SwiftUI `.sensoryFeedback`, so no in-app toggle).

## Vocabulary
| Action | Haptic |
|---|---|
| Send a message | light impact |
| Stop an agent | light impact |
| Agent turn completes **while its chat is the visible screen** | success notification |
| Approve a plan | success |
| Reject / dismiss a plan | warning |
| Submit an answer to a question | success / warning (matches whether it reached the socket) |
| Toggle mode / model / thinking level in the composer | selection |

## Design
- A tiny central `Haptics` helper (`Services/Haptics.swift`) holds the whole vocabulary in one place: `impact`, `notify`, `selection`.
- The turn-completion haptic is the **store-testable slice**: `ConversationStore` exposes an observable `visibleCompletionCount` that bumps only when a `.done` arrives for the **focused** session while `isChatVisible` is true. Background sessions, other tabs, and hidden screens never bump it. `ChatView` fires the success haptic via `.sensoryFeedback(.success, trigger:)`.
- Discrete button actions call the helper directly; composer wiring is confined to `ChatInputBar` to keep `ChatView` edits minimal.

## Scope
iOS only. New: `Services/Haptics.swift`, a test file. Modified: `Stores/ConversationStore.swift`, `Views/Chat/ChatView.swift`, `Views/Chat/ChatInputBar.swift`.

## Testing
- `cd ios && swift test` → **190 tests pass** (+5 new: visible-bumps, not-visible-no-bump, background-session-no-bump, cancelled-no-bump, accumulation).
- `xcodebuild build … -scheme HiveMobile -destination 'platform=iOS Simulator,name=iPhone 17 Pro'` → **BUILD SUCCEEDED**.

## Note
Touches `ChatView` (also edited by #306) and `ConversationStore`; edits are in disjoint regions, so this should rebase cleanly. Haptics are physical — best verified on a device (the simulator does not play them).

PR #306, opened 2026-07-06
Show a loading skeleton while conversation history loads (#290)

florian-bellotti:
Closes #290.

## What changed
The chat message area used to render a blank `Spacer()` while a first conversation's REST history loaded. It now shows a shimmering skeleton (`ConversationLoadingSkeleton` — alternating leading/trailing rounded-rect "bubbles" using the existing `.shimmer()` modifier and theme tokens), gated behind a ~300ms appearance delay so fast/cached loads never flash it. The empty-state and message-list branches and `loadMessages` are unchanged.

Skeleton shows only when a history fetch is in flight AND there are no cached messages; content renders the instant it arrives.

## Tests
- Added `cachedMessagesDrivesLoadingVersusContentState` to `ConversationStoreSessionTests` asserting the exact combos the view keys off: no cache → `cachedMessages` nil (loading branch); after `applyFetchedHistory` → returns messages (instant content).
- `cd ios && swift test` → 186 tests pass.
- `xcodebuild build -scheme HiveMobile -destination 'platform=iOS Simulator,name=iPhone 17 Pro' CODE_SIGNING_ALLOWED=NO` → BUILD SUCCEEDED.

## Manual verification
Screenshot of the loading skeleton to follow.

Issue #297, opened 2026-07-02
iOS haptic feedback for key actions


florian-bellotti:
## What to build

Sparse, purposeful haptic feedback driven by observable state via the SwiftUI sensory-feedback modifier: light impact on send/stop, success on turn completion in the visible chat, success/warning on plan approval/rejection and question submission, selection on composer toggles. Never for background sessions or non-visible screens.

## Acceptance criteria

- [ ] The mapped events (send, stop, visible turn completion, plan approve/reject, question submit, mode/model toggles) each produce the specified feedback type
- [ ] No haptics for background-session completions, other tabs, or non-visible screens; no haptics tied to stream chunks or tool events
- [ ] Feedback is driven by observable state changes via the sensory-feedback modifier; the driving state transitions are store-tested (including the visibility gate)
- [ ] System haptics setting is respected automatically; no in-app toggle
- [ ] `cd ios && swift test` passes

## Blocked by

None - can start immediately

---

## Problem Statement

Hive on iPhone is completely silent to the touch: sending a message, stopping an agent, approving a plan, or having an agent finish a long turn produces no tactile response at all. On a platform where users expect meaningful moments to be felt, the app's most significant interactions — committing work to an agent, and the agent completing that work — feel flat and unconfirmed. Users glance back at the screen to double-check that a tap registered, and the moment an agent finishes while they are watching passes without any physical acknowledgment.

## Solution

Key actions gain restrained, purposeful haptic feedback. Sending a message and stopping an agent give a light impact — a quiet confirmation that the action landed. When an agent turn completes while the user is actively viewing that chat, a success notification haptic marks the moment. Approving or rejecting a plan and submitting an answer to an agent's question give success or warning feedback matching the outcome. Toggling mode or model in the composer gives subtle selection feedback. Haptics fire only for events on the screen the user is looking at, automatically respect the system's haptics settings, and are deliberately sparse — significant moments only, never a buzz per event.

## User Stories

1. As a user sending a prompt to an agent, I want a light tap when the message sends, so that I feel the action was committed without watching for UI changes.
2. As a user stopping a running agent, I want a light tap on the stop action, so that I know the interrupt registered immediately.
3. As a user watching an agent work in chat, I want a success haptic the moment the turn completes, so that I can look away from the stream and still catch the finish.
4. As a user approving an agent's plan, I want success feedback on approval, so that the go-ahead feels deliberate and confirmed.
[...]

PR #316, opened 2026-07-07
iOS mobile app improvements (consolidated)

### LINKED ISSUE #285: iOS Hub server-unreachable state with retry

florian-bellotti:
Consolidates all the open iOS PRs into one branch so we can review, fix, and merge them together. Based directly on \`main\`; Push Notifications capability is intact (the dev-only removal for personal-team signing is **not** included here).

## What's included

| Area | Change | Issue |
|------|--------|-------|
| History | Loading skeleton while a conversation's history loads | #290 |
| Connection | Debounced connection banner, dropped the connecting toast | #304 |
| Chat | Fix disclosure scroll jump/blink + tool-open lag | — |
| Motion | Honor Reduce Motion in activity animations | #296 |
| Push | Navigate to the workspace on push-notification tap | #294 |
| Haptics | Sparse haptic feedback for key actions | #297 |
| Hub | Honest server-unreachable state with retry | #285 |
| Onboarding | First-run funnel (host/port/token + VPN/reachability hint) | — |
| Composer | Native model picker, image de-dup, clean remove-image control, model catalog retry | — |
| Navigation | Smooth bottom tab bar transition in/out of conversations | — |
| Send | Retry a failed send after reconnecting; honest disconnect banner; **resubscribe before the first send to fix "Not subscribed to workspace"** | — |

Closes #285, #290, #294, #296, #297, #304

## Supersedes
Replaces the individual PRs #306, #307, #308, #309, #310, #311, #312, #313, #314 (being closed in favor of this one).

## Verification
- \`swift test\` green (210), \`xcodebuild\` green, 0 warnings.
- Verified on simulator: onboarding funnel, connection + project load against a live backend, model picker, tab bar.
- iOS-only. No backend/frontend changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Hive can run as a local web app, a Tauri desktop app connected to a local or rem
 
 **Apps and integrations**
 - React 19 + Vite frontend with Tauri v2 desktop packaging.
-- Native SwiftUI iOS client with Brain, workspace conversations, session switching, chat, model selection, PR status, scripts, push notifications, task tracker, and Codex activity rendering.
+- Native SwiftUI iOS client with first-run onboarding, Brain, workspace conversations, session switching, chat, model selection, PR status, scripts, push notifications, task tracker, connection status banner with tap-to-reconnect, and Codex activity rendering.
 - GitHub OAuth device flow through `gh`, PR status enrichment, project `.env` management, global instructions, skills, Team agents, subagents, prompt settings, theme/accent settings, Telegram notifications, APNs, Tailscale-friendly connection settings, external terminal opening, and VS Code remote SSH opening from Tauri.
 
 ## Prerequisites
--- a/backend/src/ws/stream.test.ts
+++ b/backend/src/ws/stream.test.ts
@@ -605,6 +605,39 @@ describe("WS /ws/hub", () => {
     await endSession(wsId, dataDir).catch(() => {});
   });
 
+  it("authorizes a user_message that races the subscribe bootstrap on the same socket", async () => {
+    // Regression: a workspace event sent immediately after sync_workspaces, on
+    // the same socket, must not be rejected as "Not subscribed" just because the
+    // async bootstrap has not finished. Subscription intent is recorded before
+    // the bootstrap await, so the racing message is authorized.
+    const ws = (await app.injectWS("/ws/hub")) as WebSocket;
+    const messages: WsOutgoing[] = [];
+    ws.on("message", (data: Buffer) => {
+      const envelope = JSON.parse(data.toString()) as HubOutgoing;
+      if (!("workspaceId" in envelope)) return;
+      if (envelope.workspaceId === wsId) messages.push(envelope.event);
+    });
+
+    // Subscribe and immediately fire a workspace event, back to back, so the
+    // event frame arrives while sendWorkspaceBootstrap is still awaiting.
+    syncWorkspaces(ws, [wsId]);
+    ws.send(hubEvent(wsId, { type: "user_message", content: "race" }));
+
+    await waitForMessage(
+      messages,
+      (msgs) => msgs.some((m) => m.type === "status" && m.streaming === true),
+    );
+
+    expect(
+      messages.some(
+        (m) => m.type === "error" && m.message.includes("Not subscribed"),
+      ),
+    ).toBe(false);
+
+    ws.close();
+    await endSession(wsId, dataDir).catch(() => {});
+  });
+
   it("returns an error when user_message targets a deleted session id", async () => {
     const { session } = await getOrCreateSession(wsId, dataDir, CONV_CMD);
     const deletedSessionId = session.sessionId;
--- a/backend/src/ws/stream.ts
+++ b/backend/src/ws/stream.ts
@@ -769,13 +769,17 @@ export async function streamRoutes(app: FastifyInstance, opts: StreamRoutesOptio
             socket.send(JSON.stringify({ type: "pong" }));
           }
         } else if ("type" in parsed && parsed.type === "sync_workspaces") {
-          await handleSyncWorkspaces(
-            hub,
-            parsed.workspaceIds,
-            parsed.focusWorkspaces,
-            parsed.prWorkspaces,
-            parsed.forceBootstrap === true,
-          );
+          try {
+            await handleSyncWorkspaces(
+              hub,
+              parsed.workspaceIds,
+              parsed.focusWorkspaces,
+              parsed.prWorkspaces,
+              parsed.forceBootstrap === true,
+            );
+          } catch (err) {
+            app.log.error({ err }, "sync_workspaces handler failed");
+          }
         } else if ("workspaceId" in parsed && "event" in parsed) {
           await handleWorkspaceMessage(hub, parsed.workspaceId, parsed.event);
         }
--- a/backend/src/ws/stream.ts
+++ b/backend/src/ws/stream.ts
@@ -498,21 +498,29 @@ export async function streamRoutes(app: FastifyInstance, opts: StreamRoutesOptio
       }
     }
 
-    // Subscribe to new workspaces and bootstrap.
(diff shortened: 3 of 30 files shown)


florian-bellotti:
Added three review-hardening commits (branch review of this branch):

- **19842aa** — Wait for a fresh hub connection before retrying a failed workspace send. The send-retry loop broke on a stale `.connected` (the receive loop hadn't flipped state yet when `wsTask.send` threw), so the retry hit the dead socket. Now gated on a connection-generation counter. +1 regression test; `swift test` green.
- **c11826b** — Treat a blank `serverHost` as not configured (onboarding + banner). A type-then-clear left `serverHost=""`, which suppressed onboarding and showed a dead reconnect banner. Both gates now trim/empty-check. `swift test` green + `xcodebuild` BUILD SUCCEEDED.
- **20b9f94** — Log `sync_workspaces` handler failures instead of leaving them unhandled (single 3-line `try/catch`, minimal backend touch). typecheck + 50 WS tests + lint green.

Issue #370, opened 2026-07-27
Frontend: single connection record with a runtime access token


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 1 of 3 (`auth-token`).**

Source material: PR #342 (`install-flow-impl`) already contains a working implementation of this. Cherry-pick it — do not rewrite from scratch. #342 is a reference branch, not a base to merge.

## What to build

The desktop and web clients currently describe their server as a loose set of values shaped around Tailscale — an IP, a port, an optional SSH user — stored as separate keys, with the auth token compiled in at build time. Changing the token means rebuilding the app, and there is nowhere to record the protocol or the credentials of a server reached over its public address.

After this ticket, a client holds **one connection record** describing exactly one server: its host, port, protocol, access token, and the user accounts associated with it. Every outbound call — REST, WebSocket, PTY sockets, browser-screencast socket, authenticated image URLs — carries the token from that record. An operator can paste a token into Settings and immediately reach a token-protected server without rebuilding anything.

**The product has no users yet: write no migration code.** Do not read, convert, or preserve the old storage keys. Delete them and the code that produced them.

Two user identities belong on the record, and they are not the same: the account used to run the install over SSH (typically root), and the service account the operator should use day to day for editor and terminal sessions. Existing "open in editor / open a terminal" actions must resolve their target from this record.

## Acceptance criteria

- [ ] A single stored connection record holds host, port, protocol, access token, the day-to-day SSH user and the admin login user; it is the only source of truth for the server address
- [ ] The record is validated on read: a malformed or partial record is treated as "no server configured" rather than crashing or half-connecting
- [ ] The derived server URL omits the port when it is the protocol default, and bracket-wraps IPv6 literals
- [ ] Every outbound call carries the token: REST requests, the hub WebSocket, the PTY sockets, the browser-screencast socket, and authenticated image URLs. A caller that already set its own auth header is not overridden
- [ ] Settings shows the current server with editable host, port and token, and connecting verifies reachability before saving, reporting an unreachable server and a rejected token as distinct errors
- [ ] The editor and terminal actions resolve their host and user from the connection record
- [ ] The previous Tailscale-shaped storage and its hook are deleted, along with their tests. No migration path, no legacy key reads, no build-time token fallback
- [ ] Frontend `lint`, `typecheck` and tests are green

0xlny:
## Blocked by

- #365

Issue #368, opened 2026-07-27
Documentation: prerequisites page and rewritten setup guide


0xlny:
Part of the onboarding/install flow for the open-source release. Documentation, unblocked — can start immediately and is linked from the installer's first screen.

## What to build

The current remote-setup guide documents a single path: install Tailscale, clone the repository, build on the server, run under a process manager, and type the server address into the app. It has no public-IP path, no security guidance beyond binding to a tailnet address, and it does not match the guided installer being built.

After this ticket, a newcomer arriving at the repository can tell — before starting — whether their server qualifies and what they must prepare themselves. And the setup guide describes the actual install paths the product ships.

Two audiences, two documents. A short **prerequisites** page, linked from the installer's first screen, that an operator reads in under a minute. And the **setup guide**, rewritten around the guided installer with the manual path kept for people who prefer it.

## Acceptance criteria

- [ ] A prerequisites page states the supported operating systems and architectures, and that systemd is required
- [ ] It states that the installer connects over SSH with a key, and that it needs either a root login or an account whose `sudo` works — noting that a `sudo` password is supported and will be asked for when needed
- [ ] It explains the two network modes: reaching the server over a private tailnet, or over its public address protected by the access token. Tailscale is described as something the operator sets up themselves beforehand, on **both** the server and the machine running Hive, with a link to Tailscale's own instructions
- [ ] It states plainly what the installer will and will not change on a server that already runs other software: it adds a firewall rule only if a firewall is already active and never changes the global policy, it installs its own runtime inside its own directory rather than touching the system one, and it can be removed with a generated uninstall script
- [ ] The setup guide is rewritten around the guided installer, and the manual path is updated to match what the product actually does today — including configuring the access token
- [ ] No documentation of unfinished work outside the README backlog section
- [ ] All copy in English

0xlny:
## Review: **needs-rework — but the fault is in my sequencing, not the writing**

The reviewer found two blockers, and it was right to. It is also right for the wrong reason, and that is on me.

The documentation describes `HIVE_AUTH_TOKEN_SHA256`, `HIVE_ALLOWED_HOSTS`, `HIVE_ALLOWED_ORIGINS` and a token field in Settings > Connection. **None of those exist on `main`.** Reviewed against `main`, the guide is actively dangerous: it tells an operator to set only the digest form, which on today's backend means *no authentication at all*, and its own verification step cannot detect that — `curl` with a bearer returns 200 whether or not auth is enabled.

[...]

PR #378, opened 2026-07-27
feat(auth): runtime access token — hashed at rest, Host guard, WebSocket origin guard

### LINKED ISSUE #365: Backend: hashed-token auth, Host guard, and WebSocket origin guard
Part of the onboarding/install flow for the open-source release. **PR stack 1 of 3 (`auth-token`).**

Source material: PR #342 (`install-flow-impl`) already contains a working implementation of this. Cherry-pick it — do not rewrite from scratch. #342 is a reference branch, not a base to merge.

## What to build

Today a Hive backend with no token configured accepts every request from anyone who can reach it, and the desktop client can only carry a token baked in at build time. That makes the public-IP deployment path unusable and leaves tailnet deployments protected by network reachability alone.

After this ticket, an operator can start a backend that holds only the **SHA-256 of its access token** on disk, and every `/api/*` route and every WebSocket endpoint rejects requests that do not present the matching secret. The server also refuses requests whose `Host` header could indicate a DNS-rebinding attack, and refuses browser-initiated WebSocket upgrades from untrusted origins. The health endpoint stays open and unauthenticated so a provisioning script can poll it.

Plaintext-token configuration keeps working alongside the hashed form, so a manually configured server is unaffected.

## Acceptance criteria

- [ ] The backend accepts an auth expectation as either a plaintext token or a lowercase-hex SHA-256 digest; a request authorizes when its presented secret matches either form
- [ ] With neither form configured, the server stays open (local/dev default) and this is covered by a test
- [ ] Secrets are compared in constant time; a length mismatch cannot short-circuit the comparison
- [ ] Tokens are accepted via `Authorization: Bearer`, an `x-hive-token` header, and a `token` query parameter (the query form is required for `` and browser WebSockets, which cannot set headers)
- [ ] Every WebSocket endpoint enforces the same expectation as the REST routes
- [ ] Requests whose `Host` header is not an IP literal, `localhost`, a Tailscale MagicDNS name, or an explicitly allowed host are rejected with 403; the allowlist is configurable by environment variable
- [ ] Browser WebSocket upgrades from an origin outside the allowed set are rejected with 403; the Tauri webview origins are allowed by default, and dev-server origins are allowed only outside production
- [ ] The health endpoint bypasses the auth, Host and origin guards, and its payload carries no secrets
- [ ] Backend `lint`, `typecheck` and tests are green; new behaviour is covered by unit tests including the negative cases

0xlny:
**PR 1 of 3** in the onboarding/install stack. `feat/provisioning` will branch from this one, and `feat/tool-auth` from that, so the tip of the stack is always testable end to end.

Draft: this PR also carries #370 (the client half) once it is implemented. #365 is done and reviewed.

## Why this is separable and lands first

It fixes a real hole in `main` independently of the installer. Today `isAuthorized` returns `true` whenever no token is configured, and the desktop client can only carry a token compiled in at build time via `VITE_HIVE_AUTH_TOKEN`. There is no runtime token field anywhere in the web or desktop UI. That makes the public-address deployment path unusable, and leaves a tailnet deployment protected by network reachability alone.

## What #365 delivers

- **Hash-at-rest.** A server can hold only the SHA-256 digest of its token, so a provisioned install never stores the secret it checks against. Plaintext stays supported for manually configured servers. Presenting the digest itself does not authorize — there is a test for that, because a digest that leaks off disk must not be replayable.
- **Host guard.** Rejects DNS-rebinding attempts: a malicious site resolving its own domain to the backend carries that domain in the `Host` header. IP literals, localhost, Tailscale MagicDNS names and `HIVE_ALLOWED_HOSTS` entries are accepted.
- **WebSocket origin guard.** CORS does not cover WebSockets, so browser upgrades get their own check. Native clients send no `Origin` and stay allowed.
- `/health` is exempt from all three so a provisioning script can poll it. It carries no secrets.

## Review findings, fixed on branch

Two came out of adversarial review and both were real:

1. **Constant-time comparison was not.** `safeEqual` short-circuited on length mismatch, which the ticket explicitly forbade, leaking token length through response latency in plaintext mode. Both operands are now reduced to a fixed-width digest before comparison. **The test named for that criterion was theatre** — it asserted the boolean result, so it passed identically with or without the short-circuit. It now asserts the structural property, and was mutation-verified twice independently: reinstating the short-circuit fails 3 tests.
2. **Failed authentication was not rate limited.** Hooks ran auth-then-limit, and a 401 `reply.send()` ends the Fastify lifecycle, so the limiter never saw rejected requests. Registering it before the auth hook fixes it, covered by a test.

Also fixed: `?token=a&token=b` returned 500 instead of 401.

## Deliberate scope note

CORS moved from `origin: true` to an allowlist. This was **not** asked for by the ticket, and is kept because with reflect-any-origin and no token configured, any website could read the local API. Consequence to document in #368: running `npm run tauri dev` against a server started with `NODE_ENV=production` is now CORS-blocked; `HIVE_ALLOWED_ORIGINS` is the escape hatch.

## Verification

`backend` lint, typecheck and tests green — 102 files, **1789 tests**. The constant-time and rate-limit fixes were each mutation-tested: reverting them fails exactly the tests written for them.

Closes #365


0xlny:
## #370 landed — PR 1 is complete and ready for review

`5ed1a14 feat(connection): single server record with a runtime access token` — 34 files, +1140/−840.

### What the client half delivers

One record describes one server (host, port, protocol, token, and **two** SSH identities), and every outbound call carries the token: REST, hub socket, PTY sockets, browser-screencast socket, image URLs, and project favicons.

**Two SSH identities, deliberately.** The service account is used for editor and terminal sessions; the admin login is kept only to re-run an install. They are not interchangeable — the service account owns the repositories, so an editor session connecting as root takes ownership of every file it saves, after which the agent can no longer write them. `useSshHost` resolves `sshUser` and never `adminUser`, and a test asserts it.

**No migration.** The product has zero users, so the previous storage keys and their hook are deleted outright — not read, not converted, not deprecated. `rg VITE_HIVE_AUTH_TOKEN frontend/` and `rg useTailscaleConfig|LEGACY_KEYS|hive-tailscale-ip frontend/` both return nothing.

### Three real bugs found while implementing, and fixed

1. **Project favicons carried no token at all.** `/api/projects/:id/favicon` sits behind the auth hook and an `` cannot set headers, so every favicon silently 401'd against a protected server.
2. **The connection badge lied.** It probed `/health`, which is deliberately exempt from auth so a provisioning script can poll it before any token exists. A rejected token therefore showed a green **Connected** badge while every request failed — the badge reported the one thing working and hid the one thing broken. It now probes an authenticated endpoint and distinguishes *token rejected*, *client refused* (the new Host/origin guards return 403), *unreachable*, and *not configured*, each with its own guidance. Regression guard: *"does not report connected when only the unauthenticated health route answers"*.
3. The card's subtitle read `Connected to host:port` regardless of actual status.

### Verification

`frontend` lint, typecheck, tests green — 137 files, **1650 tests**. `backend` unchanged since the #365 commit (`git diff HEAD -- backend/` empty).

Every outbound call site was enumerated by grep and checked individually rather than summarised, and `tests/lib/auth-token-call-sites.test.ts` locks all six down.

### Known cost, stated plainly

The badge now costs one `/api/projects` request per check instead of a `/health` ping, and `AppLayout` mounts the hook app-wide, so it fires on app load. `refetchInterval` is `false`, so it is not a hot poll. If it ever matters, the honest fix is a cheap authenticated endpoint on the backend — deliberately out of scope here.

### Doc edits included

`README.md` and `GETTING_STARTED.md` lose their `VITE_HIVE_AUTH_TOKEN` rows. Not scope creep: this change deletes that mechanism, and shipping docs that describe it would be shipping documentation for something that no longer exists. The full rewrite is #368, which lands last in the stack.

Closes #370

0xlny:
Superseded by #386 (consolidated onboarding stack). Branch kept for reference.

Issue #376, opened 2026-07-27
Installer part 2: run the install, land in a working Hive


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 2 of 3 (`provisioning`).** This ticket closes the zero-terminal install: after it, the whole path from a bare server to a working Hive is demoable.

Source material: PR #342 (`install-flow-impl`). Its provisioning screen and run-management are worth cherry-picking. **Its navigation controls and its post-install behaviour are deliberately different** — read below. #342 is a reference branch, not a base to merge.

## What to build

The operator has a validated, trusted server that has not yet been touched. This ticket runs the install and lands them in a working Hive.

**Once the install starts, there is no way out but through.** No back, no cancel, no skip, no "continue later", no "start over". This is safe *because* the run resumes: progress is persisted and the script is resumable, so an operator who quits the app finds the installer where they left it. Abandonment cannot be prevented anyway — a laptop sleeps, a network drops — so the goal is not to trap anyone, it is to remove controls that promise something they cannot deliver. Note that the reference implementation's cancel button terminated only the *local* SSH process while the remote install carried on regardless.

**A failed run is not a run in progress.** On failure the operator gets retry, which resumes the idempotent install, and back, so a mistyped address or an occupied port can be corrected instead of trapping them.

**The operator can see what is happening.** The install already streams every command's output; the screen renders a step checklist and a collapsible live log. That output is never persisted to browser storage.

**On success the app becomes itself.** The generated access token and the connection are stored, the installer closes, and what remains is the ordinary app. Settings shows the one configured server and offers re-running the installer to reinstall or move servers. There is **no disconnect action** — nothing may put the app back into an unconfigured state on purpose.

## Acceptance criteria

- [ ] The install runs over the established connection and renders a live step checklist reflecting each step's state
- [ ] A collapsible log pane shows the streamed command output as it arrives; no log content is written to browser storage
- [ ] While a run is in progress there is no back, cancel, skip, continue-later or start-over control anywhere on the screen
- [ ] Quitting and relaunching the app returns to the in-progress install and resumes it rather than restarting from the beginning
- [ ] A run is not started twice if the screen remounts, and re-entering an in-progress run attaches to it rather than launching another
- [ ] On failure, a typed error with an actionable hint is shown, alongside retry and back; retry resumes rather than restarting
- [ ] On success the access token reported by the install and the server address are stored as the connection, the installer state is cleared, and the ordinary app opens
[...]

Issue #375, opened 2026-07-27
Installer part 1: from launch to a validated, trusted server


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 2 of 3 (`provisioning`).**

Source material: PR #342 (`install-flow-impl`). Its wizard shell, persistence and screen chrome are worth cherry-picking. **Its screen sequence is deliberately different** — read below. #342 is a reference branch, not a base to merge.

## What to build

Today, launching the packaged desktop app without a configured server produces a broken app: it loads, fails every request against its own asset origin, and shows a warning that retries forever. There is no onboarding, no redirect, no call to action. It only appears to work in development because a dev proxy hides it.

After this ticket, launching the app with no server configured opens an installer instead. This ticket covers everything up to the moment the install begins — after it, the operator has a validated, trusted connection to a server that has not yet been touched.

**The installer is a layer in front of the app, and the only "not configured" state there is.** Do not build a reduced or disconnected variant of the app. Once a server is configured, the app is exactly what it is today.

**The first screen is the only exit.** It offers two paths: install Hive on a server, or connect to a Hive server that already exists — the second being an inline form for address, port and token. That second path *is* the skip; there is no separate skip control.

**The network mode is declared by the operator, never inferred.** Hive does not install or configure the private network — that is a prerequisite the operator arranges beforehand, on both the server and this machine, and the prerequisites documentation covers it. The choice therefore selects **how the server will be exposed**, not how the app connects: reachable only over the private network, or reachable at its public address and protected by the access token. The address the operator types is the address that reaches the server right now, and it does not change later.

**Nothing may fail during the install that could have been caught here.** The connect step runs the preflight over the already-open SSH connection and reports its findings. Anything fixable — a busy port, an unwritable directory, a missing private-network interface — is reported here, where the operator can correct the form, rather than mid-install.

## Acceptance criteria

- [ ] Launching the desktop shell with no server configured opens the installer. The web build is unaffected. No reduced or disconnected variant of the app exists
- [ ] The first screen offers installing on a server or connecting to an existing one; the latter is an inline address, port and token form that configures the app and dismisses the installer
- [ ] The first screen links to the prerequisites documentation
- [ ] A network screen collects the exposure mode, the address, the port, and — behind an "advanced" disclosure — the install and data directories, pre-filled with defaults
- [ ] The port defaults to the project's production port, not its development port, everywhere it appears
[...]

Issue #373, opened 2026-07-27
Tauri SSH sidecar: connect, trust, preflight, stream the install, three privilege modes


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 2 of 3 (`provisioning`).**

Source material: PR #342 (`install-flow-impl`) contains a working SSH sidecar whose transport, host-key trust and streaming design should be kept wholesale. **Its private-network key handling is deleted here and its privilege handling is extended.** #342 is a reference branch, not a base to merge.

## What to build

The desktop app must be able to run the provisioning script on the operator's server itself — no terminal, no copy-paste. This ticket is the bridge: the desktop shell connects over SSH, establishes trust in the server's identity, runs the preflight, then runs the install and streams its progress into the UI as it happens.

Two properties from the reference implementation are worth stating because they are easy to lose. The script is **never written to the server's disk** — it is streamed to a shell reading from standard input — and the server's host key is trusted in an **app-private** store rather than the operator's own known-hosts file, so Hive never edits the operator's SSH configuration. Both must survive.

**Privilege escalation is extended.** The reference implementation supports only a root login or an account whose escalation needs no password, which excludes exactly the hardened, already-in-use servers this flow targets. A password must be supported. The mechanism is settled: the escalation tool can read a password from the head of standard input, one byte at a time, and leaves the remainder intact for the shell that follows — this was verified against its source and at the syscall level, so prefixing the password to the streamed payload is sound. It is only sound **when the password is actually consumed**: if escalation happens to need no password, the line is not read and becomes the first line of the script. Preflight already establishes which case applies, so the password line is sent only when it is known to be needed, and any cached credential is invalidated first so the check cannot be bypassed at the last moment.

The password must never appear in a process argument, on disk, or in persisted client state — and must be stripped from captured error output before it reaches the UI.

## Acceptance criteria

- [ ] The operator's usable SSH keys can be listed for selection, excluding files that are not keys, and indicating which are unusable because they are passphrase-protected and not loaded in an agent
- [ ] Connecting to a host retrieves its key, prefers a modern key type, and surfaces a fingerprint for the operator to approve before anything else happens
- [ ] Approved host keys are stored in an app-private known-hosts file, never the operator's own. Subsequent connections verify strictly; a changed key aborts with a distinct error rather than silently re-trusting
- [ ] Preflight can be invoked over the same connection and its findings returned as structured data
- [ ] The provisioning script is streamed over standard input and never written to the server's filesystem
[...]

Issue #371, opened 2026-07-27
provision.sh: preflight, configurable paths, non-intrusive firewall, service-account SSH access, uninstall


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 2 of 3 (`provisioning`).**

Source material: PR #342 (`install-flow-impl`). **Its firewall step and its hardcoded paths are deliberately superseded here** — read the acceptance criteria before copying. #342 is a reference branch, not a base to merge.

## What to build

The guiding assumption for this whole flow is that the operator's server is **already doing something else**. It is rarely a clean box dedicated to Hive. The installer from the previous ticket still behaves as if it owns the machine: it insists on a pristine server, writes to fixed locations, and cannot be undone.

After this ticket the installer is a good guest. It can report what it would do without changing anything. It puts its files where the operator tells it to. It touches the machine's firewall as little as possible. It leaves the operator able to work in their code from an editor without breaking file ownership. And it can be removed.

**On the firewall, the reference implementation is actively dangerous and must not be reproduced.** It sets a default-deny inbound policy, allows only port 22, and force-enables the firewall. On a server where the firewall was installed but inactive, that single command severs every service without an explicit rule — a web server, a database, anything — and on a server whose SSH listens on a non-standard port it severs the very session running the install, with no way to undo it. A survey of nine widely used installers (Tailscale, Docker, k3s, Netdata, Coolify, Dokploy, rustup, Nix, Homebrew) found **not one** that touches the firewall; the universal convention is to document the ports and leave the policy to the operator. Since the access token is now the security boundary, the firewall is defence in depth here, not the mechanism.

**On removal**, follow the k3s convention: generate the uninstall script at install time, parameterised with the paths actually used, so it removes what was really installed even when the operator chose non-default locations.

**On editor access**, note that the service runs as a dedicated unprivileged account and owns the repositories and worktrees. An operator whose editor connects as root will silently take ownership of every file they save, after which the agent can no longer write them.

## Acceptance criteria

- [ ] A preflight mode runs the checks and reports findings **without modifying anything**, exiting cleanly whatever it finds
- [ ] Preflight reports: operating system and architecture support, whether the target port is free, whether the chosen directories' parents exist and are writable with sufficient free space, whether Hive is already installed, whether the firewall is currently active, whether privilege escalation works without a password, and — when the tailnet network mode is requested — **whether the tailnet interface actually exists**
[...]

PR #379, opened 2026-07-27
feat(install): provisioning — release pipeline, provision.sh, SSH sidecar, installer

0xlny:
**PR 2 of 3** in the onboarding/install stack. Based on #378, not on `main` — review its diff against `feat/auth-token`. `feat/tool-auth` will branch from this one, so the tip of the stack stays testable end to end.

Draft. Carries #366 (landed), then #369, #371, #373, #375, #376.

## #366 — release pipeline (landed)

Tagging a release publishes a self-contained backend tarball per architecture with a checksum alongside. Asset names derive from version and architecture alone, so the installer builds the URL without querying the API and verifies the download before extracting.

The tarball ships prebuilt native modules, which is exactly why the runtime version is pinned — they are compiled against a specific ABI and a server on a different major cannot load them. This is the constraint that drives the private-runtime decision in #369.

**Verified further than the ticket asked.** Rather than trusting the build script's own assertions, the verifier unpacked the produced tarball, loaded both native modules independently, booted the backend from the unpacked tree and got a live answer from `/api/projects`.

All nine third-party action pins were resolved against the GitHub API — including dereferencing annotated tag objects — and every one matches the tag named in its trailing comment. A wrong pin is worse than no pin. `ci.yml` was pinned to the SHAs behind the tags **currently in use** rather than bumped, so a major upgrade is not smuggled in under a pinning fix.

### Known gap

*"Verified end to end on a prerelease tag"* is not satisfied and could not be: until this branch exists on the remote, `release.yml` is not on a branch Actions will run. Now that it is pushed, this is the first thing to exercise before PR 2 leaves draft.


0xlny:
## All six tickets landed — PR 2 is complete

`#366` release pipeline · `#369` provision.sh · `#371` shared-VPS safety · `#373` SSH sidecar · `#375` installer part 1 · `#376` installer part 2

### The central claim is now proven rather than asserted

The premise of this whole flow is that the operator's server **already runs something else**. A containerised lane installs Hive onto a machine already serving a web server and a database:

- firewall **inactive** → ports 80 and 5432 still answer an outside peer afterwards, and the firewall is still inactive with zero rules
- firewall **active** → only the backend port is added, the `Default:` policy line is **byte-identical**, port 80 still reachable

A preflight run leaves 791 paths and 172 service entries unchanged. Non-default install and data directories are honoured end to end. The generated uninstaller removes the install and leaves the data.

For contrast, the approach this replaces set a default-deny policy, allowed only port 22, and force-enabled the firewall — which on that same machine severs every service without an explicit rule, and on a server with SSH off port 22 severs the session running the install.

### Five real defects found by writing the lanes rather than assuming

1. **systemd's start limit** made a rollback report itself as unrecoverable, though the rollback was correct.
2. **The install directory was created before the marker claiming it** — a crash in between left an install that could never be resumed, because the next run read its own directory as a foreign occupant.
3. **The prior branch's `build.sh` claimed truncation safety in a comment but never implemented it.** Here the bundle is one function invoked on its last line, so a short `curl` read is a syntax error that executes nothing — and the streamed payload was diffed byte-for-byte against the published bundler output.
4. **The end-to-end lane could not run unattended at all**: its published checksum could not be rewritten when the build container ran as root.
5. **That lane's cleanup silently failed** against root-owned build output, leaking ~680 MB per run — a disk-fill bug had it ever reached CI.

### Two guards beyond the specification, kept

The token digest is read back off disk and pattern-checked, and a final step then calls the API **unauthenticated and stops the service** unless it is refused. An empty expectation makes the backend open, so a silently-failed token generation would otherwise boot a wide-open server on all interfaces with no error.

### Deliberately different from the prior approach

No Tailscale install and no auth key — a private network is a prerequisite the operator arranges. No system-wide runtime: Node lives inside Hive's own directory, checksum-pinned, and the system one is never read or modified. No stripped-down app when nothing is configured — the installer is a layer over the unchanged app. No cancel while running, because the prior implementation's cancel killed only the local SSH process while the remote install carried on.

### Verification boundary

Backend 1789, frontend 1705, `cargo test` 24, provision contract 107 assertions, shellcheck clean, six containerised lanes green. **No run against a real VPS and no Tauri desktop build** — that is Lenny's end-to-end pass. Also unverified: arm64, Debian 12/13, and real `firewalld`/`nftables` hosts.

Two gaps found and filed rather than quietly left: #381 (a rebuilt server cannot be re-trusted) and #382 (SSH on a non-default port).

0xlny:
Superseded by #386 (consolidated onboarding stack). Branch kept for reference.

Issue #377, opened 2026-07-27
Installer final screen: connect accounts, then open Hive


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 3 of 3 (`tool-auth`).** Final ticket: after it, a newcomer goes from an empty desktop app to a Hive that can actually run an agent, without ever opening a terminal.

## What to build

A freshly installed server can run, but no accounts are connected, so no agent can do any work. The installer currently ends there. This ticket adds the last screen: connect your accounts, then open Hive.

The screen embeds the same tool panel that Settings uses — one component, two placements.

**It never blocks.** The install has already succeeded and the server is running, so the rule that forbids navigation during the install does not apply here. Every sign-in is optional and anything skipped remains available in Settings. Connecting **either** Claude or Codex is enough for Hive to be usable, and the screen says so rather than implying all three are required.

## Acceptance criteria

- [ ] The installer's final screen embeds the tool panel and offers sign-in for GitHub and the agent providers
- [ ] Continuing is always possible regardless of what is connected; there is no gate and no required sign-in
- [ ] The screen states that connecting one agent provider is enough, and reflects live what is currently connected
- [ ] Anything skipped is reachable afterwards from Settings, and doing it there produces the same result
- [ ] Finishing closes the installer and opens the ordinary app
- [ ] Frontend `lint`, `typecheck` and tests are green
- [ ] Verified end to end: from a bare server, a newcomer following only the documentation reaches a Hive that runs an agent, with no terminal at any point

## Blocked by

- #374
- #376

Issue #374, opened 2026-07-27
Connect agent and GitHub accounts from the UI


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 3 of 3 (`tool-auth`).**

Source material: PR #342 (`install-flow-impl`). **Its GitHub sign-in is deliberately discarded** — read below. Its Claude sign-in should be cherry-picked. #342 is a reference branch, not a base to merge.

## What to build

A provisioned server has the agent CLIs installed but no accounts connected, so it cannot do any work. This ticket connects them from the UI, with no terminal.

The bar for each is the flow people already know: press connect, a browser opens, confirm a code, done.

**GitHub already works this way in this codebase and must be reused, not rebuilt.** Hive already implements the GitHub device flow natively against the GitHub API — it requests a device code, shows the code and URL, polls for the token, and then configures both the GitHub CLI and git's credential helper on the server. The reference branch reimplemented this by spawning the GitHub CLI in a pseudo-terminal and regular-expression-parsing its human-readable output to recover the same code. That is strictly worse: it breaks whenever the CLI changes a label or its colours, and it needed shell fixtures to be testable at all. **Delete that implementation and wire the UI to the existing native flow.** Confirm the requested scopes cover what the CLI and git operations need.

**Claude has no HTTP equivalent** — its sign-in command requires an interactive terminal, opens a browser, and expects an authorization code pasted back. Driving it through a pseudo-terminal is unavoidable and the reference implementation does it correctly: validate the recovered token's shape before storing, write it with owner-only permissions, and never place it in a process argument. Keep that.

**Codex follows whatever #367 concluded.** Do not re-investigate; implement the approach that spike recommends.

**Signing in to either Claude or Codex is enough.** Hive treats providers as interchangeable, so nothing here may block on having both, or on having any particular one. The UI states what is connected and what that enables; it never gates.

## Acceptance criteria

- [ ] GitHub connects through the codebase's existing native device flow: press connect, get a code and a URL, confirm in the browser, and the panel reflects the connected account
- [ ] Completing GitHub sign-in leaves both the GitHub CLI and git's credential helper configured on the server, and a failure to configure git credentials is reported rather than silently ignored
- [ ] The pseudo-terminal GitHub sign-in from the reference branch is deleted along with its tests and shell fixtures
- [ ] Claude connects through its sign-in command driven over a pseudo-terminal: the browser opens, the operator pastes the authorization code, and the panel reflects the connected account
[...]

Issue #372, opened 2026-07-27
Tool detection, install and update: API and Settings panel


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 3 of 3 (`tool-auth`).**

Source material: PR #342 (`install-flow-impl`) already contains a working implementation of this. Cherry-pick it — do not rewrite from scratch. #342 is a reference branch, not a base to merge.

## What to build

Hive drives agent CLIs installed on the server. The provisioning flow installs them, but nothing lets an operator see their state afterwards or update them without opening a terminal — which is the one thing this whole effort exists to avoid.

After this ticket, Settings shows a panel listing each tool Hive depends on, whether it is installed, and which version. Tools that are missing or outdated can be installed or updated from that panel, with progress and failures surfaced in the UI rather than swallowed.

The operations are **durable and resumable**: an install that takes minutes must survive the operator navigating away or the app reconnecting, and must not start twice if two clients ask at once. Failures carry a typed reason — a network problem reads differently from a command that ran and failed — and a truncated tail of the real output, so an operator has something actionable rather than "it failed".

The panel is a self-contained component. A later ticket embeds the same component in the installer's final screen, so it must not assume it is rendered inside Settings.

## Acceptance criteria

- [ ] An API reports, for each tool Hive depends on, whether it is installed, its version, and whether it is authenticated
- [ ] Install and update operations run server-side, are individually addressable, and report progress and terminal state to any client that asks
- [ ] Operations are serialised: a second request for an operation already running joins it rather than starting a duplicate
- [ ] Operation state survives a client disconnecting and reconnecting, and a restart of the server does not leave an operation wedged in a permanently running state
- [ ] Failures carry a typed reason distinguishing at minimum a network failure from a command failure, plus a bounded excerpt of the real output
- [ ] Installs place tools under the service account's own directory rather than system-wide
- [ ] The panel renders tool state and exposes install and update actions, and is usable both from Settings and, later, from the installer, without assuming its container
- [ ] No secret or command output is persisted to browser storage
- [ ] Backend and frontend `lint`, `typecheck` and tests are green, including the failure paths

## Blocked by

- #370

Issue #367, opened 2026-07-27
Spike: can Codex sign-in complete on a headless remote server?


0xlny:
Part of the onboarding/install flow for the open-source release. **PR stack 3 of 3 (`tool-auth`).** This is a spike, not a feature.

## What to build

The guided setup signs the operator into their agent CLIs on the remote server, from the desktop UI, with no terminal. GitHub and Claude have known-good approaches. Codex does not: `codex login` opens a browser and expects the OAuth callback to reach `localhost:1455` — but on a remote install the browser runs on the operator's laptop while the callback must arrive at the *server's* loopback. It is unknown whether this can complete unassisted.

This ticket produces a decision, not a shipped feature: establish what actually happens, and what the fallback is.

Hive already treats providers as optional and interchangeable — signing in to **either** Claude or Codex is enough to make the product usable. So a negative result here is acceptable and does not block the release; it only changes what the Codex card does.

## Acceptance criteria

- [ ] Attempt a real `codex login` on a headless remote Linux server driven from a desktop machine, and record exactly where it succeeds or fails
- [ ] Determine whether the CLI can be pointed at a callback the operator's machine can reach, or whether an SSH tunnel is required, or whether an API-key path is the only viable option
- [ ] Check whether the CLI offers a non-interactive or device-code style flow in its current version, using its own documentation as the source of truth rather than assumption
- [ ] Write the verdict and the recommended approach into the issue, with the commands and versions used
- [ ] If no unassisted flow exists, state plainly what the Codex card should do instead (for example: offer an API key field, or direct the operator to run one command over SSH) so the dependent ticket can be implemented without re-investigating

0xlny:
## Verdict: solved. Codex has a device-code flow, and it needs no terminal.

Tested on Ubuntu 24.04 (KVM guest), `codex-cli 0.145.0`.

`codex login --device-auth` exists and is exactly the flow we wanted. **There is no `localhost:1455` callback involved, no SSH tunnel, and no pseudo-terminal.** It behaves like the GitHub device flow:

```
Follow these steps to sign in with ChatGPT using device code authorization:

1. Open this link in your browser and sign in to your account
   https://auth.openai.com/codex/device

2. Enter this one-time code (expires in 15 minutes)
   TZNS-9ZBD2
```

Then it polls until the operator confirms in their browser.

### Key result: it works with no TTY

[...]

PR #380, opened 2026-07-27
feat(setup): connect the agent and GitHub accounts from the app

0xlny:
**PR 3 of 3** in the onboarding/install stack. Based on #379, which is based on #378 — review its diff against `feat/provisioning`. The tip of this branch is the full flow end to end.

Draft. Carries #372 (landed), then #374, #377, and #368 last.

## #372 — tool detection, install and update (landed)

Settings lists each tool Hive depends on with its version and sign-in state, and can install or update the ones it manages. Operations are **durable**: keyed by tool so a second request joins rather than duplicates, written to disk before the request returns, and reaped on construction so a restart mid-install surfaces as a retryable failure instead of a spinner that never resolves.

### The brief was wrong and the agent said so

I instructed the implementer to take the boot reaper from PR #342. **There isn't one.** That branch's runner carries an explicit comment — *"Operations intentionally do not survive backend restarts"* — and the only trace of the idea is an unused error code. The durability layer here was built from scratch because the ticket requires it.

Worth recording for the rest of this stack: **#342's description overstates its code.** It advertises "durable resumable operations with boot reaper"; the code says the opposite. Treat that PR as reference material to read, not as a specification to trust.

### Deliberate departures, all defensible

- **`gh` is reported but not installable from the panel.** Its only trustworthy path is the checksum-pinned tarball in the provisioning script; re-deriving those digests in TypeScript would put pinned hashes in two languages. The panel says so rather than offering a button whose result cannot be verified.
- **Claude installs from npm, not `curl | bash`.** The provisioning script installs the npm package and the panel reads the registry for "update available" — mixing the two would make the offer incoherent with what the update does.
- **Tool-operation failure reasons live in a new module, not the provisioning error taxonomy.** The provision contract test asserts exact set equality between the shared taxonomy and the bash one; adding `network`/`command_failed` there would break it, and reusing an existing code would mean labelling an unreachable registry an install failure.
- **The narrower CLI-status endpoint is deleted.** The new one is a strict superset and the page was repointed, which would have left the old route with zero consumers. Verified: no remaining reference anywhere, including iOS.

### Fixed along the way

A failed state write inside the background task surfaced as an unhandled rejection, which would have taken the process down. Covered by a test that runs an operation to success against an unwritable data directory.

### Verification

Root `lint`, `typecheck`, `knip` clean. **backend 1842, frontend 1653**, 82 of them new. Provision contract suite still green.

Not verified: no install or update was executed against a live server, and the sign-in probe command shapes are inherited from #342 rather than confirmed against the current CLIs — #374 will confirm them.


0xlny:
## All tickets landed — the stack is complete

`#372` tool detection/install · `#374` account sign-in · `#377` installer final screen · `#368` documentation

The tip of this branch is the whole flow: a bare Ubuntu or Debian server to a working Hive that can run an agent, without a terminal at any point.

### Three defects found by verifying rather than assuming

1. **The installer would have unmounted itself one screen from the end.** It was gated on the absence of a stored connection, but the connection is stored *before* the accounts screen renders — so the screen that makes the server usable could never have appeared. Now latched open, closing only on its own signal.
2. **A stored Claude credential would never have reached an agent run.** The workspace environment strips that whole variable prefix, deliberately, to avoid leaking an ambient token from whatever launched the backend. Detection assumed otherwise, so the sign-in would have been decorative. Now injected explicitly for the provider that needs it and dropped for the one that authenticates elsewhere.
3. **Claude's authorisation URL is a hyperlink escape whose visible label wraps to the terminal width.** Recovering it by stripping escapes and matching a URL — which is what the prior branch did — yields only the first fragment: a URL that parses cleanly and resolves to nothing.

### Codex turned out simpler than assumed

The spike (#367) was run on a real headless box. `codex login --device-auth` needs no pseudo-terminal and no loopback callback — it prints a URL and a code to standard output and polls. But it **deletes any existing credential the moment it starts**, so an abandoned or expired attempt would log an operator out of a working Codex. Backed up before, restored on any non-connected outcome, recovered at startup if the process died between, and confirmed before disturbing a working sign-in.

### The GitHub scope set was missing `workflow`

GitHub gates it separately from `repo` and rejects any push whose branch touches a workflow file. Hive agents edit repositories, so this was a live failure with a confusing error. `delete_repo` is still deliberately absent — asking every operator to grant repository deletion to service a rare cleanup path is the worse trade, and that cleanup already reports its own failure.

### Documentation

Every claim was checked against the source. Doing so surfaced a real gap that predates this work and is now filed as #384: the shipped pm2 production profile binds every interface with no access token, and the old guide's verification step returned 200 whether or not authentication was enabled — so following it confirmed a server that was in fact open.

### Verification boundary

Backend 1909, frontend 1735, `cargo test` 24, provision contract 107 assertions, six containerised lanes, all links and anchors resolved. **No run against a real VPS and no Tauri desktop build** — that is Lenny's end-to-end pass, which is the one acceptance criterion nothing here can satisfy.

0xlny:
Superseded by #386, which consolidates the whole onboarding stack (#378 → #379 → #380) into one PR. Branch kept for reference.

PR #435, opened 2026-08-10
feat: one-button server update from the app

0xlny:
## Summary

Iteration B of the update flow: the app can now update the backend itself. With PR #433's version surfaces plus this, the full loop is: desktop auto-updates → app detects the differing server → one click runs `provision.sh --update` over SSH.

### Ordering (the structural decision)
The provisioner is compiled into the desktop binary at the app's version, so **the app updates first, the server second** — the bundled script always installs the app's own version. The UI encodes this: while an app update is pending, the server button is hidden (a run already in flight keeps its progress visible). After the relaunch, a once-per-launch toast nudges toward Settings > Updates. No persisted orchestration state: mismatch detection is the orchestrator.

### What's in
- **Rust sidecar**: `ProvisionOptions.update` flag → sends `--update` alone (the script rejects `--allowed-host`/`--ssh-public-key` in that mode and reads identity from the install manifest). Privilege preflight, streaming, and redaction reuse the install path unchanged.
- **SSH identity**: the installer persists the key *path* on the stored connection (never key material); pre-existing connections get a one-time key picker. The sudo password is prompted per run only on `SSH_PASSWORD_REQUIRED` and never stored. ConnectionSettings carries the install-owned identity (`adminUser`, `sshKeyPath`) through a reconnect only when the host is unchanged.
- **`server-update.ts`**: small module store (idle/running/failed/done), same pattern as the desktop-update store. The mismatch rule is one shared helper, `serverVersionDiffersFromApp`: any comparable difference offers convergence to the app's version — the button names its target, so a downgrade is a visible, supported choice.
- **Updates page**: converge button ("Update server to X"), inline step progress, failure + retry with the manual command as fallback, and a warning dialog when workspaces are busy — counted from the **live WS status map** (status transitions never invalidate the projects query, whose snapshot could miss an agent started from another client; the Brain counts too).
- **Single-server semantics**: `switchServer` refuses while a server update is provisioning — the sidecar cannot be cancelled, and a run's state must never describe a server the client no longer points at. Desktop `install()` and the provisioning run also guard against each other (relaunch would kill the SSH run; the Rust `RunGuard` serializes sidecar runs).
- Server-side rollback/health-check untouched — the provisioner already owns correctness there.
- Also carries a small unrelated fix: React plugin aligned with Vite 8.

### Review & audit
Adversarially reviewed by agent (2 real bugs + 2 risks fixed and re-verified), then a 3-pass external audit, now closed: live busy source, refuse-switch-during-run, host-scoped SSH identity, and the helper rename all came out of it. Accepted cosmetic nits: transient "Server updated." flash, "workspace" wording for a busy Brain, case-sensitive host comparison.

## Testing
- `frontend`: lint, typecheck, full suite — 1807 tests, covering the store, the page flow (busy dialog, key picker, password retry, fallback command), the boot prompt, both cross-guards, and the switch refusal.
- Rust: new `script_args` update-mode test; `cargo` unavailable on this Linux host (GTK deps) — CI's macOS build covers it.
- Live e2e validated on an OrbStack VM: fresh install → mismatch toast → one-click update with local tarball sideload, including the sudo-password dialog path.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -109,8 +109,10 @@ the generated uninstall script are all covered in **[Prerequisites](prerequisite
 
 ### Update the backend
 
-Wait for active agents, terminals, and automations to finish, then run the provisioner from the
-exact release you want:
+The desktop app offers this flow from Settings > Updates once the app itself is current: it runs
+the same update over SSH, warns when agents are still running, and asks for the escalation
+password only when the server requires one. To run it by hand instead, wait for active agents,
+terminals, and automations to finish, then run the provisioner from the exact release you want:
 
 ```bash
 curl -fsSL https://github.com/unfence-labs/hive/releases/download/v0.1.0-beta.2/provision.sh \
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -109,10 +109,11 @@ the generated uninstall script are all covered in **[Prerequisites](prerequisite
 
 ### Update the backend
 
-The desktop app offers this flow from Settings > Updates once the app itself is current: it runs
-the same update over SSH, warns when agents are still running, and asks for the escalation
-password only when the server requires one. To run it by hand instead, wait for active agents,
-terminals, and automations to finish, then run the provisioner from the exact release you want:
+The desktop app offers this flow from Settings > Updates: it converges the server to the app's
+version, whether that means an upgrade or a visible downgrade. It runs the update over SSH, warns
+when agents are still running, and asks for the escalation password only when the server requires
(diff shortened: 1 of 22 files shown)


0xlny:
External audit round (3 passes, closed): the busy-workspace warning now counts from the live WS status map instead of the stale projects query; switchServer refuses while a server update is provisioning (single-server semantics — no multi-server store by design); the install-owned SSH identity is dropped when the connection host changes; the mismatch rule is one shared helper renamed serverVersionDiffersFromApp, with the convergence policy (app is the authority, downgrade is a visible choice) stated in code and docs. Full frontend suite 1807/1807, lint + typecheck clean.

PR #495, opened 2026-08-27
fix(ios): make wide markdown tables readable in chat

0xlny:
## Problem

Wide GFM tables in agent responses render unreadably on iOS: MarkdownUI's stock `gitHub` table style applies `.fixedSize(horizontal: false, vertical: true)` to the table grid, forcing it into the bubble width and collapsing columns down to one character per line.

## Fix

Two theme overrides in `.whisperChat` (verified against swift-markdown-ui 2.4.1 source):

- **`.table`**: wrap the grid in a horizontal `ScrollView` with `.fixedSize(horizontal: true, vertical: true)` so columns take their natural width, the same idiom the theme already uses for code blocks. Border and alternating-row colors are re-supplied from `WhisperColor` tokens since the override replaces the gitHub block style.
- **`.tableCell`**: gitHub cell style (semibold header row, padding, line spacing) plus `FontSize(.em(12/14))` to match inline code, and a small custom `Layout` (`TableCellWidthCap`) that caps cell width at 260pt so prose cells wrap at a readable measure while short columns (`#`, `Risk`, ...) stay compact. A `Layout` rather than `frame(maxWidth:)` because a flexible frame only clamps the size it reports and never proposes the cap to its child, so the grid measured row heights as if every cell were a single line and wrapped cells overlapped the rows below (caught during on-device testing). The Layout proposes the cap during measurement, so a cell's reported height is its wrapped height. Cells also keep their natural width inside the column, so GFM `:--:` / `--:` column alignments keep working.

Matches web behavior (streamdown's `table-wrapper` already horizontal-scrolls with natural widths) and the standard pattern in ChatGPT iOS / GitHub mobile.

Also adds a wide Findings-style table message to the `#Preview` for quick visual verification.

## Testing

- `swift test` not run: no Swift toolchain on the Linux dev host, and the SPM target excludes `Views/` anyway; this file only compiles under Xcode.
- First commit verified on device: natural column widths and horizontal scroll work; it exposed the row-height overlap fixed by the second commit, whose device re-test is pending.

### DIFF
--- a/ios/HiveMobile.xcodeproj/project.pbxproj
+++ b/ios/HiveMobile.xcodeproj/project.pbxproj
@@ -334,8 +334,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/gonzalezreal/swift-markdown-ui";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 2.4.0;
+				kind = exactVersion;
+				version = 2.4.1;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/ios/HiveMobile/Views/Chat/MessageBubble.swift
+++ b/ios/HiveMobile/Views/Chat/MessageBubble.swift
@@ -1156,6 +1156,36 @@ extension Theme {
             .background(WhisperColor.codeBlockBg, in: RoundedRectangle(cornerRadius: 8))
             .markdownMargin(top: .em(0.4), bottom: .em(0.4))
         }
+        // ── Tables ──
+        // The stock table style squeezes columns into the bubble width, which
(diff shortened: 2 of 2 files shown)


0xlny:
Follow-up commit 1897b498: on-device testing showed wrapped cells overlapping the rows below. `frame(maxWidth:)` only clamps the size a cell reports and never proposes the cap to its child, so the grid measured row heights as if every cell were single-line. Replaced with a small `Layout` (`TableCellWidthCap`) that proposes the cap during measurement, so reported heights are wrapped heights. Side effect: the GFM column-alignment tradeoff noted in the description is gone — cells keep their natural width, so `:--:` / `--:` columns align correctly again.

